PR #2, opened 2015-09-01
Make client compatible with change to dockerclient api

wilstrup:
The dockerclient api has been updated with a new mandatory parameter - I propose to just default to force=false
Added support for registry authentication

### DIFF
--- a/container/client.go
+++ b/container/client.go
@@ -153,7 +153,7 @@ func (client dockerClient) IsContainerStale(c Container) (bool, error) {
 func (client dockerClient) RemoveImage(c Container) error {
 	imageID := c.ImageID()
 	log.Infof("Removing image %s", imageID)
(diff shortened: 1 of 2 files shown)


bdehamer:
@wilstrup I was just reviewing this and see now that you closed it. Is this change no longer needed?

wilstrup:
@bdehamer It's because I realized that the authentication are on a per-registry basis. My commit tried to apply the authentication to all registries. So I'm rethinking the solution.

bdehamer:
ok, that's the same issue I ran up against . . . and why I punted on the authentication stuff for the initial release :smile: 

There is already a file that the Docker CLI uses to store those creds (when you do a `docker login`). I was sort of thinking of trying to leverage that in some way, but didn't put much more thought into it than that.

PR #9, opened 2016-02-03
Configure with ENV vars and allow skipping of restart

haswalt:
This allows us to use ENV vars to configure most options. In the case when using an orchestration framework access to changing the command isn't viable however ENV vars can be easily utilised to configure the container.

In the same scenario not restarting allows the orchestration tools to handle starting the container itself.

### DIFF
--- a/main.go
+++ b/main.go
@@ -23,6 +23,7 @@ var (
 	client       container.Client
 	pollInterval time.Duration
 	cleanup      bool
+	noRestart    bool
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -49,17 +50,25 @@ func main() {
...
+			Name:   "no-pull",
+			Usage:  "do not pull new images",
+			EnvVar: "WATCHTOWER_NO_PULL",
 		},
 		cli.BoolFlag{
-			Name:  "cleanup",
-			Usage: "remove old images after updating",
+			Name:   "no-restart",
+			Usage:  "do not restart containers",
+			EnvVar: "WATCHTOWER_NO_PULL",
```
mithrandi: This environment variable name is wrong; probably you wanted something like `WATCHTOWER_NO_RESTART`.

haswalt:
thanks @mithrandi fixed it

haswalt:
Any chance this can be merged?

PR #13, opened 2016-02-24
Allow private image pulls

frodopwns:
This PR allows you to pass docker hub credentials to the watchtower image as ENV vars and thus find images that are private.

### DIFF
--- a/container/client.go
+++ b/container/client.go
@@ -111,7 +111,7 @@ func (client dockerClient) StartContainer(c Container) error {
 
 	log.Infof("Starting %s", name)
 
-	newContainerID, err := client.api.CreateContainer(config, name)
+	newContainerID, err := client.api.CreateContainer(config, name, nil)
 	if err != nil {
 		return err
 	}
@@ -153,7 +153,7 @@ func (client dockerClient) IsContainerStale(c Container) (bool, error) {
 func (client dockerClient) RemoveImage(c Container) error {
 	imageID := c.ImageID()
 	log.Infof("Removing image %s", imageID)
-	_, err := client.api.RemoveImage(imageID)
+	_, err := client.api.RemoveImage(imageID, true)
 	return err
 }
 
--- a/container/client.go
+++ b/container/client.go
@@ -13,6 +13,12 @@ const (
 	defaultStopSignal = "SIGTERM"
 )
 
+var auth = dockerclient.AuthConfig{
+	Username: "",
+	Password: "",
+	Email:    "",
+}
+
 // A Filter is a prototype for a function that can be used to filter the
 // results from a call to the ListContainers() method on the Client.
 type Filter func(Container) bool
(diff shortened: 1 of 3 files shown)


frodopwns:
Looks like some functions in the tests need to be updated to reflect newly required parameters.

eg. RemoveImage() now requires *AuthConfig or nil

edit: on second thought this may be testing with the wrong version of dockerclient

frodopwns:
Also i guess the change to ubuntu is a drastic one.  I did this because the watchtower binary would not run in scratch or alpine....possibly due to dynamic linking?

phumpal:
@frodopwns very cool. 

I've tested this and it seems to work well with Docker Hub but doesn't seem to support images pulled from a private (self-hosted) repo.

frodopwns:
@phumpal yeah it wasn't readily apparent to me in the https://github.com/samalba/dockerclient how you would pull from another registry.

bahlo:
:+1:

nafg:
What's the status of this?

bahlo:
@frodopwns: I checked and I think, you need to set the `DockerClient.URL` property to change the registry. I've not tested this, but it is used in all registry operations.

I'm not sure though ¯\_(ツ)_/¯

container/client.go
```diff
@@ -111,7 +116,19 @@ func (client dockerClient) StartContainer(c Container) error {
...
-	newContainerID, err := client.api.CreateContainer(config, name)
+	var err error
+	var newContainerID string
+	if username != "" && password != "" && email != "" {
+		auth := dockerclient.AuthConfig{
+			Username: username,
+			Password: password,
+			Email:    email,
+		}
+		newContainerID, err = client.api.CreateContainer(config, name, &auth)
```
bahlo, line 127: Maybe set `dockerclient.URL` manually to `os.Getenv("REPO_REGISTRY")`, if not empty, or something?

frodopwns:
Looking into using other than Docker Hub repos with credentials...

chrisferry:
What's the status of this?  Any chance it will support ECR Auth?

rossvz:
Is there a timeline on this being merged?

curry684:
The switch to Ubuntu makes this pretty much worthless for any serious use given the amount of disk space it's killing now if you don't use Ubuntu:14.04 in other containers. The point of a watchtower should pretty much be sitting unobtrusively in a very small (5MB) corner of the system methinks.

frodopwns:
@curry684 that part probably isn't necessary. I just didn't want to use that previous image in my testing for some reason i can't remember.  This is still just a go app so it should run anywhere go binaries do.

curry684:
I agree, but it still got merged.

frodopwns:
@curry684 https://github.com/CenturyLinkLabs/watchtower/pull/21

fdattein:
@frodopwns funny enough, changes in #21 aren't in master, despite github saying it has been merged. Am I missing something?

PR #15, opened 2016-03-15
Pass in HostConfig to create operation

jamiehannaford:
This pull request fixes three problems:
1. HostConfig information is now sent in the Create call, rather than the Start call. This is in line with the latest  Docker Remote API (v1.22), which has [deprecated payloads](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#start-a-container) in the Start call.
2. The `CreateContainer` function in dockerclient now accepts a [third argument for auth config](https://github.com/samalba/dockerclient/blob/master/interface.go#L16).
3. The `RemoveImage` function in dockerclient now accepts a [second bool argument](https://github.com/samalba/dockerclient/blob/master/interface.go#L50).

Apart from not being able to build, another bug is that HostConfig information is not adhered to when containers are redeployed (e.g. their port mappings).


jamiehannaford:
@bdehamer Hey, would you mind taking a look at this if you have any free time? Thanks!

stffabi:
Resolved by #26. This PR could be closed.

PR #39, opened 2017-01-01
Simple Scheduling for Upgrades (Maintenance Window)

### LINKED ISSUE #12: Suggestion: maintenance window for updates
Some of us without high availability service setups would prefer to keep our restarts / disruptions within scheduled maintenance windows. Being able to specify a time range to do updates in (or equivalently, a time range to _not_ do updates in) would be helpful in this regard.

stffabi:
Possibility to define a cron expression which specifies when to check for updated images. This allows to have a schedule in which updates should be made and therefore one could define a maintenance window.

Fixes #12


stffabi:
Builds as soon as #37 is merged.

Issue #23, opened 2016-09-12
Multiple network support


dolanor:
I use docker-compose, and I have a container connected to 2 networks, but only 1 is connected again during an upgrade.
Without this it's hard to have tiered application (which became quite simple with docker-compose)

nuth:
Hi, this is a issue for me as well. It seems watchtower doesn't connect to the correct network at all, just the default.

Technifocal:
Identical issue here.

dolanor:
For me watchtower connects to the first network I believe. Need to test more.

Also, it doesn't create links back.

A before/after diff:

``` diff
--- wp.before	2016-11-23 00:54:20.623839414 +0100
+++ wp.after	2016-11-23 00:56:14.348854650 +0100
@@ -1,7 +1,7 @@
 [
     {
-        "Id": "e7e60b52cb31bf3e98adf9390b556e2092af41ed82bb3a23dc40d9519ba6b714",
-        "Created": "2016-11-22T23:52:45.235164032Z",
+        "Id": "316a6c8809ef994eb482b2ed6d23a90dfec7e993f6f69f39ffdf2b7e9d864006",
+        "Created": "2016-11-22T23:55:36.739528722Z",
         "Path": "/entrypoint.sh",
         "Args": [
             "apache2-foreground"
@@ -13,17 +13,17 @@
             "Restarting": false,
             "OOMKilled": false,
             "Dead": false,
-            "Pid": 25541,
+            "Pid": 25979,
             "ExitCode": 0,
             "Error": "",
-            "StartedAt": "2016-11-22T23:52:53.955561665Z",
+            "StartedAt": "2016-11-22T23:55:37.527022118Z",
             "FinishedAt": "0001-01-01T00:00:00Z"
         },
-        "Image": "sha256:d71a6f7248d2caff6a9c39af9e1dc35fce6276d1754a7780830156342abca887",
-        "ResolvConfPath": "/var/lib/docker/containers/e7e60b52cb31bf3e98adf9390b556e2092af41ed82bb3a23dc40d9519ba6b714/resolv.conf",
-        "HostnamePath": "/var/lib/docker/containers/e7e60b52cb31bf3e98adf9390b556e2092af41ed82bb3a23dc40d9519ba6b714/hostname",
-        "HostsPath": "/var/lib/docker/containers/e7e60b52cb31bf3e98adf9390b556e2092af41ed82bb3a23dc40d9519ba6b714/hosts",
[...]

PR #40, opened 2017-01-02
Fix Multiple Network Support #23

dolanor:
Following the discussion on #23 and the patch from @rosscado in #26 (which greatly helped me), I think I fixed the problem watchtower had with restarting a container previously connected to multiple networks.
I can say that It works here :tm: with docker compose.

Hope I'll help other people too.

### DIFF
--- a/container/client.go
+++ b/container/client.go
@@ -147,7 +147,7 @@ func (client dockerClient) StartContainer(c Container) error {
 		return err
 	}
 
-	for k, _ := range simpleNetworkConfig.EndpointsConfig {
+	for k := range simpleNetworkConfig.EndpointsConfig {
 		err = client.api.NetworkDisconnect(bg, k, creation.ID, true)
 		if err != nil {
 			return err
--- a/container/client.go
+++ b/container/client.go
@@ -8,6 +8,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 	dockerclient "github.com/docker/docker/client"
 	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/network"
 	"golang.org/x/net/context"
 )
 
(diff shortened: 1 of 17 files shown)


Poeschl:
Any news on this?

stffabi:
This will be merged soon. I'm currently refactoring the overall build process. As soon as this is finished, new updated images will be published.

stffabi (approved):

stffabi:
Fixes #23

stffabi:
Thanks @dolanor for your contribution.

container/client.go
```diff
@@ -118,17 +119,49 @@ func (client dockerClient) StartContainer(c Container) error {
...
-	creation, err := client.api.ContainerCreate(bg, config, hostConfig, nil, name)
+	creation, err := client.api.ContainerCreate(bg, config, hostConfig, simpleNetworkConfig, name)
 	if err != nil {
 		return err
 	}
 
 	log.Debugf("Starting container %s (%s)", name, creation.ID)
 
-	return client.api.ContainerStart(bg, creation.ID, types.ContainerStartOptions{})
+	err = client.api.ContainerStart(bg, creation.ID, types.ContainerStartOptions{})
```
stffabi, line 145: @dolanor I've just seen that you reattach the networks after the container has already been started. I think it would be better if we first reattach the networks and afterwards start the container again. Or did you have some problems and made it in this order?

dolanor: The explanation of this behaviour is detailed in the git message :wink: 

https://github.com/v2tec/watchtower/pull/40/commits/f8a2f80b92da3beae08efd84d2f024cefe738da8

stffabi: Yes, I'm aware of the behaviour that we can't directly create the container with all attached networks. But my question was more if we could do it this way.

1. ContainerCreate(simpleNetworkConfig)
2. NetworkDisconnect(simpleNetworkConfig)
3. NetworkConnect(networkConfig.EndpointsConfig)
4. ContainerStart()

I think starting the container with only one attached network and afterwards disconnect it and reconnecting the correct ones, may result in strange behaviors for the containers.
It seems like [docker-compose](https://github.com/docker/docker/issues/29265#issuecomment-265909198) also first creates the container, then connects/disconnect the networks and as last step starts the container.

dolanor: Oh I see.
Good point. In fact I didn't understand the answer from thaJetzah like that. But you're right.
So I guess my current implementation doesn't work like docker-compose then.
I will fix it tomorrow.

stffabi: Thanks, that would be great.

PR #55, opened 2017-01-26
set default registry server-addres for Docker HUB images

rreinurm:
Current Docker HUB users can't pull private DockerHUB images without specifying docker image with server path like "index.docker.io/rreinurm/private-image". But in docker-compose and docker commands don't require to explicitly specify DockerHUB registry-address for private images, I would just use imagename "rreinurm/private-image".

This pull request appends "index.docker.io" as a default registry-server address to the docker image name to be able to pull them with same ux as docker commands.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 watchtower
 vendor
 .glide
+debug
+.vscode
--- a/container/container.go
+++ b/container/container.go
@@ -61,6 +61,10 @@ func (c Container) ImageName() string {
 		imageName = fmt.Sprintf("%s:latest", imageName)
 	}
 
+	if strings.Count(imageName, "/") == 1 {
+		imageName = strings.Join([]string{"index.docker.io/", imageName}, "")
+	}
+
 	return imageName
 }
 


stffabi:
Thanks @rreinurm for your contribution.
This really seems to be an issue, but I would prefer not to fix the images onto `index.docker.io` at this location. I think we should left these image names untouched and rather fallback to the default registry in `trust.go`, if no registry is defined in the image. That's the place where the login credentials are determined. Furthermore accordingly to this comment https://github.com/docker/docker/blob/e9c0de0de65182915d1cd205f06f5c903c20a29f/cli/command/registry.go#L24 the default registry may depend on the platform. As a result we should also call `/info` to determine the default registry, similarly as in the linked `registry.go`.

@rreinurm would it be possible for you to change the implementation of the PR?

stffabi (changes requested):
See previous comment.

rreinurm:
sure, I will refactor my implementation

stffabi:
That's great, thanks.

stffabi:
@rreinurm just wanted to ask back if you did find some spare time to change your implementation?

wmbutler:
Seems like the easiest way to handle this would be to grab the value off the `~/.docker/config.json` file since it's already present.

```json
{
        "auths": {
                "https://index.docker.io/v1/": {
                        "auth": "d21idXRsZXI6dzRjb3gyaXY="
                }
        }
}
```

wmbutler:
Seems like it would be easiest to grab this off `~/.docker/config.json` since it's already present:

```
{
        "auths": {
                "https://index.docker.io/v1/": {
                        "auth": "lkfjljfrfjrferofjeo"
                }
        }
}
```

morpheus65535:
@stffabi Is there an ETA for this to be include in a release?

simonaronsson:
As @rreinurm seems to be MiA and this issue isn't something I'd say is prioritized, i'll go ahead and close this PR. If this where to be changed in the future, the path @stffabi and @wmbutler where steering against would likely be the preferred solution.

Thanks.

PR #60, opened 2017-02-07
Add ps style filtering of containers

ubergesundheit:
This pull request adds a `--ps-filter` command line flag which allows to filter containers just like the `--filter` flag of `docker ps`.

The new flag allows to only run watchtower on containers specified by it. This includes:
```
          - exited=<int> an exit code of <int>
          - label=<key> or label=<key>=<value>
          - status=(created|restarting|running|paused|exited|dead)
          - name=<string> a container's name
          - id=<ID> a container's ID
          - before=(<container-name>|<container-id>)
          - since=(<container-name>|<container-id>)
          - ancestor=(<image-name>[:tag]|<image-id>| ⟨image@digest⟩) - containers created from an image or a descendant.
          - volume=(<volume-name>|<mount-point-destination>)
          - network=(<network-name>|<network-id>) - containers connected to the provided network
```
(from `man docker ps`)


zoni:
It looks like commit 97d108f accidentally included an unrelated `actions/notify_slack.go `.

ubergesundheit:
Oh yes, I'll remove it from this pull request

PR #61, opened 2017-02-07
add slack notification cli flag

ubergesundheit:
This pull request adds sending of notifications to slack channels.

Please review the glide.yaml and glide.lock files

### DIFF
--- a/actions/notify_slack.go
+++ b/actions/notify_slack.go
@@ -1,0 +1,92 @@
+package actions
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/mozillazg/request"
+)
+
+// SlackNotifier is used to make the https requests to a specified slack
+// webhook URL
+type SlackNotifier struct {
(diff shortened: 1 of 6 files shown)


rdamazio:
This app already uses logrus which can be hooked to slack - why not just use that? (unfortunately that doesn't work for email, since logrus email integration only sends errors)

Nightbr:
Hey, the idea is pretty cool, I would really enjoy slack notification for watchtower.
If you still want to contribute on this feature, look at amazing work of @rdamazio here https://github.com/v2tec/watchtower/pull/106 and you can try to implement Slack notification with logrus and this integration https://github.com/johntdyer/slackrus

stffabi:
Yeah, would be great to have this included in the same release as the email notifications.

ubergesundheit:
I'll see if I am able to change everything to use logrus/slackrus

ubergesundheit:
See #113

PR #70, opened 2017-03-16
Update README.md

wmbutler:
Provide more clarity for private repos and add a docker-compose example.


stffabi (approved):
Thanks for your contribution.

Issue #69, opened 2017-03-16
Suggestion: Allow timeout value to be changed


Robotex:
**[Premise]**
Some of my containers require to shutdown gracefully because they have long tasks that might run for minutes before they can exit.
I haven't learned Go yet but reading your code especially the line at [update.go#L13](https://github.com/v2tec/watchtower/blob/master/actions/update.go#L13) I assume that it waits 10 seconds before forcefully killing the container in [client.go#L106](https://github.com/v2tec/watchtower/blob/master/container/client.go#L106)
**[Actual request]**
Would be it possible to add a parameter for allowing to change the timeout?
As an alternative I think just signalling the container without terminating it still through a parameter like `--no-kill` would work for me.
Thanks in advance.

stffabi:
Adding a timeout parameter would be a good suggestion. Adding a no-kill option would be a little bit harder, because if the old container doesn't get removed or renamed one can't start the new one with the same old name.

stffabi:
Will be available with the next version.

PR #74, opened 2017-03-24
Add timeout parameter

Robotex:
These changes add the timeout parameter for #69 
I've just learnt the basics of Go from a quick read of their Playground tutorial, so I apologize for any mistake

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ docker run --rm v2tec/watchtower --help
 * `--interval, -i` Poll interval (in seconds). This value controls how frequently watchtower will poll for new images. Defaults to 300 seconds (5 minutes).
 * `--schedule, -s` [Cron expression](https://godoc.org/github.com/robfig/cron#hdr-CRON_Expression_Format) in 6 fields (rather than the traditional 5) which defines when and how often to check for new images. Either `--interval` or the schedule expression could be defined, but not both. An example: `--schedule "0 0 4 * * *" ` 
 * `--no-pull` Do not pull new images. When this flag is specified, watchtower will not attempt to pull new images from the registry. Instead it will only monitor the local image cache for changes. Use this option if you are building new images directly on the Docker host without pushing them to a registry.
+* `--stop-timeout` Timeout before the container is forcefully stopped. When set, this option will change the default (`10s`) wait time to the given value. An example: `--stop-timeout 30s` will set the timeout to 30 seconds.
 * `--label-enable` Watch containers where the `com.centurylinklabs.watchtower.enable` label is set to true.
 * `--cleanup` Remove old images after updating. When this flag is specified, watchtower will remove the old image after restarting a container with a new image. Use this option to prevent the accumulation of orphaned images on your system as containers are updated.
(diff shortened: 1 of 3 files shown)


bradleyfalzon:
I'm not a maintainer, but the cli package does have a DurationFlag (https://godoc.org/github.com/urfave/cli#DurationFlag) which may be a better option so you don't need to do the cast to time.Duration.

main.go
```diff
@@ -78,6 +79,12 @@ func main() {
 			Usage:  "use TLS and verify the remote",
 			EnvVar: "DOCKER_TLS_VERIFY",
 		},
+		cli.IntFlag{
+			Name:	"timeout",
+			Usage:	"timeout before container is forcefully stopped",
```
schnapster: I suggest saying "timeout **in seconds** before container is forcefully stopped", to be more clear

stffabi:
@Robotex thanks for your contribution. I like the idea of @bradleyfalzon to use the duration flag and we should also add some documentation on the README.md.

Would it be possible for you to make these changes?

Robotex:
I have made and committed the requested changes, however ATM I don't have the build environment set up so I cannot test them. Feel free to review and comment if you encounter any compilation errors.

README.md
```diff
@@ -95,6 +95,7 @@ docker run --rm v2tec/watchtower --help
 * `--interval, -i` Poll interval (in seconds). This value controls how frequently watchtower will poll for new images. Defaults to 300 seconds (5 minutes).
 * `--schedule, -s` [Cron expression](https://godoc.org/github.com/robfig/cron#hdr-CRON_Expression_Format) in 6 fields (rather than the traditional 5) which defines when and how often to check for new images. Either `--interval` or the schedule expression could be defined, but not both. An example: `--schedule "0 0 4 * * *" ` 
 * `--no-pull` Do not pull new images. When this flag is specified, watchtower will not attempt to pull new images from the registry. Instead it will only monitor the local image cache for changes. Use this option if you are building new images directly on the Docker host without pushing them to a registry.
+* `--timeout` Timeout before the container is forcefully stopped. When set, this option will change the default (`10s`) wait time to the given value. An example: `--timeout 30s` will set the timeout to 30 seconds.
```
stffabi: What would you think about naming the option `--stop-timeout`? This makes it more clear what timeout this is.

main.go
```diff
@@ -178,6 +185,7 @@ func before(c *cli.Context) error {
 
 	cleanup = c.GlobalBool("cleanup")
 	noRestart = c.GlobalBool("no-restart")
+	timeout = c.GlobalDuration("timeout")
```
stffabi: Timeout could be negative, we should check for that and print out an error.

stffabi (changes requested):
Thanks @Robotex for your work and adjustments. I have only two little comments.

Would be great if you could take a look into those.

stffabi (approved):

stffabi:
LGTM, thanks for your contribution. I'll just squash your commits and then merge it in.

reloxx13:
pls check this: #195

--stop-timeout not found in armhf-0.3

Issue #102, opened 2017-09-06
Is it possible to configure which image/containers should be updated?


patakijv:
If considering implementing watchtower on a server where there are many containers running, we may only want certain containers to be updated based on their image getting updated locally or via a pull, not all of them across the board. Is it possible to have a configuration file or have something on the container/image itself like a label or environment variable that signals to watchtower whether it should be updated? That way we don't have to force all containers to be updated if that is not desired.

snow3461:
read again the README.md (https://github.com/v2tec/watchtower#arguments)

patakijv:
Ah ya, I overlooked the arguments that can be passed at command line. If I plan to use docker-compose to run watchtower, can those arguments be added to the 'command' key?

stffabi:
The possiblity to watch containers with an assigned label has been introduced with #104

Issue #95, opened 2017-07-18
[Idea] Ignoring containers


ToeiRei:
Watchtower makes it possible to monitor specific containers and update them - I would need it the other way round as it should i.e. leave the management containers alone but care for the (very dynamic) rest.

stffabi:
The possiblity to watch containers with an assigned label has been introduced with #104

Issue #58, opened 2017-02-02
add new container into the monitor list


zhenyulin:
As in the mode of monitoring particular containers for automatic updates and to avoid the unexpected updates of containers such as database, currently I need to stop and remove watchtower and re-run the setup script to add a new container to the monitor list.

Is there some command like "docker exec -it watchtower add <container_name>" for doing this?

stffabi:
Currently this is not possible. But there are plans to bring some kind of label select functionality into watchtower, like mentioned in #4. With such a functionality, watchtower would automatically update all containers with a specific defined label. So you could just start containers with that label and watchtower would automatically pick them up for updating.

stffabi:
The possiblity to watch containers with an assigned label has been introduced with #104

PR #104, opened 2017-10-17
Add a method of enabling or disabling containers using labels

belak:
This would potentially be a better way of handling #102, #95, #58, and #4.

I haven't tested this yet and there may be a cleaner way of doing this, but it seems like a relatively simple concept.

### DIFF
--- a/main.go
+++ b/main.go
@@ -80,6 +80,11 @@ func main() {
 			EnvVar: "DOCKER_TLS_VERIFY",
 		},
 		cli.BoolFlag{
+			Name:   "label-enable",
+			Usage:  "watch containers where the com.centurylinklabs.watchtower.enable label is true",
+			EnvVar: "WATCHTOWER_LABEL_ENABLE",
+		},
+		cli.BoolFlag{
 			Name:  "debug",
 			Usage: "enable debug mode with verbose logging",
 		},
@@ -146,7 +151,7 @@ func before(c *cli.Context) error {
 		return err
 	}
 
-	client = container.NewClient(!c.GlobalBool("no-pull"))
+	client = container.NewClient(!c.GlobalBool("no-pull"), c.GlobalBool("label-enable"))
 	notifier = notifications.NewNotifier(c)
 
 	return nil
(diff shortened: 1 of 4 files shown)


belak:
This adds two features:

* A label (`com.centurylinklabs.watchtower.enable`) which lets users specify if a container should be enabled or not on a per-container basis.
* A command line option, `--no-enable` which will let watchtower start up without watching any specific containers (unless they're specifically enabled with labels)

voglster:
+1 this, was going to write my own adding functionality similar to this until I saw this PR.  for now I'm testing belak's branch.

main.go
```diff
@@ -114,7 +119,7 @@ func before(c *cli.Context) error {
 		return err
 	}
 
-	client = container.NewClient(!c.GlobalBool("no-pull"))
+	client = container.NewClient(!c.GlobalBool("no-pull"), !c.GlobalBool("no-watch"))
```
chriswue: The command line option is "no-enable" while this looks for "no-watch". In fact I'd probably prefer a slightly more verbose option name like "no-watch-by-default"

chriswue (changes requested):
There is an option name bug which needs fixing but otherwise the changes look good to me.

Nightbr:
Any news on this, it's a real killer feature needed for watchtower!

Add to that, I think we should rename the flag as `--label-enable` in order to only take labeled containers. It makes more sense for me. What do you think?

stffabi:
@belak thanks for your contribution and thanks @chriswue for reviewing it.

I like the idea of @Nightbr to rename the flag to `--label-enable` and then only take labeled containers. I think this would be much more intuitive for the user.

On the other hand having a possibility to exclude is also handy. Maybe we could implement an include and exclude option for named lables, e.g. `--include-label mylabel` and `--exclude-label myExcludeLabel`. Specifying no include labels means all containers, and one can exclude some others based on exclude labels.

And we should also include the new command line parameter and the documentation to the readme.md

stffabi:
Any opinions how we should proceed? Should we go the `include` `exclude` way, or do you prefer some other handling?

belak:
Sorry, I've been busy with other stuff lately. At least for me, the simplest way to handle this would be `--label-enable` and that would be good enough for me. A full include/exclude seems like it might be overcomplicating this.

Also, do you have any instructions on how to run this manually for testing? Do you just use it outside a docker container manually?

main.go
```diff
@@ -79,39 +79,44 @@ func main() {
 			Usage:  "use TLS and verify the remote",
 			EnvVar: "DOCKER_TLS_VERIFY",
 		},
+		cli.BoolFlag{
+			Name:   "label-enable",
+			Usage:  "decide which containers to watch based on a label",
```
stffabi: Could you add which label we use here?

stffabi (changes requested):

stffabi:
@belak I'm fine with that 😄. Would it be possible for you to add the corresponding documentation in the readme.md? And I've put in another minor comment.

Apart from that LGTM, and I'm looking forward to merge this.

stffabi:
For testing I always use it outside of a docker container. That works quite well.

belak:
I'm open to any updates in wording... this was just thrown together based on the other documentation.

stffabi:
LGTM, thanks again for your contribution. I'm going to squash the commits and will then merge it in.

stffabi (approved):

PR #113, opened 2017-11-27
Slack notifications with slackrus

ubergesundheit:
This pull request implements and #61 with slackrus.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 9ddd729b207d71ce16ae9a0282ac87a046b9161ac4f15b108e86a03367172516
-updated: 2017-01-24T20:45:43.1914053+01:00
+hash: c0e0c69383b9b665e92c40a5ea69632ea846a645f44654411137c75c5ef5017c
+updated: 2017-11-27T11:34:21.283697431+01:00
 imports:
 - name: github.com/Azure/go-ansiterm
   version: 388960b655244e76e24c75f48631564eaefade62
@@ -25,7 +25,7 @@ imports:
   - registry/storage/cache/memory
   - uuid
 - name: github.com/docker/docker
-  version: 49bf474f9ed7ce7143a59d1964ff7b7fd9b52178
+  version: 092cba3727bb9b4a2f0e922cd6c0f93ea270e363
   subpackages:
   - api
   - api/server/httputils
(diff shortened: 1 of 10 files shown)


main.go
```diff
@@ -114,6 +114,17 @@ func main() {
...
 			EnvVar: "WATCHTOWER_NOTIFICATION_EMAIL_SERVER_PASSWORD",
 		},
+		cli.StringFlag{
+			Name:   "notification-slack-hook-url",
+			Usage:  "The Slack Hook URL to send notifications to",
+			EnvVar: "WATCHTOWER_NOTIFICATION_SLACK_HOOK_URL",
+		},
+		cli.StringFlag{
+			Name:   "notification-slack-identifier",
+			Usage:  "A string which will be used to identify the messages comming from this watchtower instance. Default if omitted is \"watchtower\"",
```
datashaman: comming -> coming

glide.lock
```diff
@@ -135,10 +138,11 @@ imports:
 - name: github.com/robfig/cron
   version: 9585fd555638e77bba25f25db5c44b41f264aeb7
 - name: github.com/Sirupsen/logrus
```
stffabi, line 141: We should remove this old logrus entry with the uppercase "Sirupsen".

notifications/slack.go
```diff
@@ -0,0 +1,34 @@
+package notifications
+
+import (
+	"github.com/johntdyer/slackrus"
+	"github.com/sirupsen/logrus"
```
stffabi: logrus should only be imported once.

stffabi (changes requested):
Thanks so much for your contribution. I've two minor comments, apart from that LGTM.

stffabi:
One more thing, could you also add some documentation? That would be great.

glide.lock
ubergesundheit: It seems at least docker and go-ansiterm are using the uppercase version of "Sirupsen".
Removing it manually the uppercase entry is re-added after running `glide update`

ubergesundheit:
So what should we do about the duplicate entry of `Sirupsen`/`sirupsen`?

notifications/slack.go
```diff
@@ -0,0 +1,33 @@
...
+type slackTypeNotifier struct {
+	slackrus.SlackrusHook
+}
+
+func newSlackNotifier(c *cli.Context) typeNotifier {
+	n := &slackTypeNotifier{
+		SlackrusHook: slackrus.SlackrusHook{
+			HookURL:        c.GlobalString("notification-slack-hook-url"),
+			Username:       c.GlobalString("notification-slack-identifier"),
+			AcceptedLevels: slackrus.LevelThreshold(log.DebugLevel),
```
rdamazio: Note that with DebugLevel you'll potentially get a lot of output - e.g. every time it checks for updates, even if there are none. Should it at least be configurable?

stffabi: @rdamazio do you know if the AcceptedLevels are independent of [setLogLevel](https://github.com/v2tec/watchtower/blob/master/main.go#L131)? So basically even if we have a non Debug-loglevel, will a Hook with threshold `DebugLevel` be triggered with Debug notifications?

Nightbr:
hey, thanks for you're work! Any news on this, really need the slack notification for watchtower a must have I think.

For the uppercase of sirupsen, if you look at logrus repo, they use sirupsen without uppercase in latest version. So I thin we should go with `sirupsen` in watchtower code.

stffabi:
@Nightbr I hope to find some time this week to review the open PRs for `v0.3.0`.

stffabi (approved):
LGTM, I've squashed some of your commits and made the nofitication level a global setting for all notifications types.

Thanks for your great contribution.

ubergesundheit:
Very nice! Thanks

Nightbr:
Thanks!

PR #116, opened 2017-12-06
github/v2tec/watchtower#114 adding if statement to accept the oneshot…

thelamer:
I am not expecting this to get merged, but this is the code I used to create a working updater I needed that will run once and exit. I do not know Go so I just wrapped the option in an if statement and ripped the core execution into and else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,16 @@ docker run -d \
 
 In the example above, watchtower will only monitor the containers named "nginx" and "redis" for updates -- all of the other running containers will be ignored.
 
+If you do not want watchtower to run as a daemon you can pass a oneshot flag and remove the watchtower container after it's execution.
+
+```bash
+docker run --rm \
+-v /var/run/docker.sock:/var/run/docker.sock \
+v2tec/watchtower --oneshot nginx redis
+```
+
+In the example above, watchtower will execute an upgrade attempt on the containers named "nginx" and "redis". Using this mode will enable debugging output showing all actions performed as usage is intended for interactive users. Once the attempt is completed, the container will exit and remove itself due to the "--rm" flag.
+
 When no arguments are specified, watchtower will monitor all running containers.
 
 ### Options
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -90,6 +100,7 @@ docker run --rm v2tec/watchtower --help
 ```
 
 * `--host, -h` Docker daemon socket to connect to. Defaults to "unix:///var/run/docker.sock" but can be pointed at a remote Docker host by specifying a TCP endpoint as "tcp://hostname:port". The host value can also be provided by setting the `DOCKER_HOST` environment variable.
+* `--oneshot` Run an update attempt against a container name list one time immediately and exit.
```
stffabi, line 103: I think we could use `--interval 0` as a special case for using the `oneshot` mode. Watchtower already has a lot of different flags and some combinations wouldn't make sense, e.g. `--oneshot` and `--interval` therefore we could just use an interval of 0 which indicates no interval.

main.go
```diff
@@ -159,53 +163,61 @@ func before(c *cli.Context) error {
 
 func start(c *cli.Context) error {
 	names := c.Args()
+        if c.GlobalBool("oneshot") {
+		log.SetLevel(log.DebugLevel)
```
stffabi, line 167: We shouldn't fix here to debug-level, we should also honor the `--debug` flag in this case. The debug flag is used in the `before` method, so you could simply drop `log.SetLevel(log.DebugLevel)` here.

main.go
```diff
@@ -159,53 +163,61 @@ func before(c *cli.Context) error {
...
+		log.Info("OneShot run: " + time.Now().String())
+	        notifier.StartNotification()
+		if err := actions.Update(client, names, cleanup, noRestart); err != nil {
+			log.Println(err)
+		}
+		notifier.SendNotification()
+        } else {
+		if err := actions.CheckPrereqs(client, cleanup); err != nil {
```
stffabi, line 175: `actions.CheckPrereqs` checks for multiple watchtower instances and stops them. I think we should also check if other watchtowers are running in the `oneshot` mode, if multiple ones are running we should simply exit the current execution and not stop the other containers.

main.go
```diff
@@ -159,53 +163,61 @@ func before(c *cli.Context) error {
...
 func start(c *cli.Context) error {
 	names := c.Args()
+        if c.GlobalBool("oneshot") {
+		log.SetLevel(log.DebugLevel)
+		log.Info("OneShot run: " + time.Now().String())
+	        notifier.StartNotification()
+		if err := actions.Update(client, names, cleanup, noRestart); err != nil {
+			log.Println(err)
+		}
+		notifier.SendNotification()
```
stffabi, line 173: It would be nice to put line 169-173 into a method, so we could also call that method on line 189 instead of having duplicated code there.

stffabi (changes requested):
Thanks for your contribution. I have placed some minor comments, would it be possible to take a look into those?

thelamer:
@stffabi sorry for the extended delay on responding here, I must admit golang is certainly not my area of expertise. I added an if statement in an area I could understand. 
I feel like you spent a lot of time trying to teach me and I appreciate that, but I would be more comfortable if you made the changes as you saw fit to support interval 0 upgrades.

simonaronsson:
I’ll take a look at it and see if its something we could incorporate in a fitting way. Thanks for your PR!

main.go
simonaronsson: I'm more inclined to one shot watchtowers skipping CheckPrereqs altogether.

simonaronsson:
Renamed it to run-once and made some small edits. Other than that, great commit. Merged! 👍

PR #137, opened 2017-12-29
added example for Cron expression

cron410:
This has been the number one question people ask after I recommend Watchtower.


stffabi:
Thanks so much for you contribution.

stffabi (approved):

Issue #112, opened 2017-11-22
Updating images for stopped containers


Umkus:
Is it possible to explicitly update specific images, not currently running containers?
The use case is docker container being used as a cron task and not running all the time.

stffabi:
Currently this is not possible. What would like to achieve? Would you like to have stopped container to be recreated with the updated images and just leave it stopped?

Umkus:
That or simply pulling newer images would be fine.

breandan:
It would be great if this worked not only for stopped containers, but all images returned by `docker images`.

darinrogers:
Another use case is when your package manager fails silently and an app with a missing dependency gets deployed. Now the container is dead because its command won't run. Even after rebuilding and creating a new, healthy image, watchtower won't restart the stopped container with the new image. 

If it instead pulled the newer image and restarted the container with it, the dead host would suddenly come back to life.

simonaronsson:
I can see this having a lot of hidden odd corner cases that needs to be addressed for it to work properly, especially to not make it interfer with the normal happy path. As such, it would likely require some groundwork beforehand. If you got the time and would like to have a stab at it, feel free. I'm currently preoccupied with getting the backlog back to some kind of manageable state.

I also believe we need to ask ourselves; is this really the right direction for watchtower? I'm not sure to be honest. Feel free to weigh in.

Thanks for creating an issue though! 🙏

Issue #76, opened 2017-03-27
Question: How to update non-running containers?


pro100coder:
I have volume container with data and it non-running. wathtower does't listing updates for it

amq:
Another use-case: an ephemeral container for running migrations

simonaronsson:
with the next release, you'll be able to update these using `--include-stopped`

PR #152, opened 2018-01-10
Implement support for updating stopped containers

aldahick:
This PR adds a command line flag, "--check-all". If true, the behavior requested by #112 is used (i.e., stopped containers are updated but not restarted). If false, the previous behavior is used (ignore stopped containers entirely).
I attempted to run the tests (with `go test`, as specified in `CONTRIBUTING.md`) but Go couldn't find any test files. I've manually tested the changes and they seem to work well.

### DIFF
--- a/container/client.go
+++ b/container/client.go
@@ -37,31 +37,41 @@ type Client interface {
 //  * DOCKER_HOST			the docker-engine host to send api requests to
 //  * DOCKER_TLS_VERIFY		whether to verify tls certificates
 //  * DOCKER_API_VERSION	the minimum docker api version to work with
-func NewClient(pullImages, enableLabel bool) Client {
+func NewClient(pullImages, enableLabel bool, checkAll bool) Client {
 	cli, err := dockerclient.NewEnvClient()
 
 	if err != nil {
 		log.Fatalf("Error instantiating Docker client: %s", err)
 	}
 
-	return dockerClient{api: cli, pullImages: pullImages, enableLabel: enableLabel}
+	return dockerClient{
+		api:         cli,
+		pullImages:  pullImages,
+		enableLabel: enableLabel,
+		checkAll:    checkAll}
 }
 
 type dockerClient struct {
 	api         *dockerclient.Client
 	pullImages  bool
 	enableLabel bool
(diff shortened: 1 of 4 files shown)


container/client.go
```diff
@@ -37,31 +37,41 @@ type Client interface {
...
+		log.Debug("Retrieving containers")
+	} else {
+		log.Debug("Retrieving running containers")
+	}
 
 	runningContainers, err := client.api.ContainerList(
 		bg,
-		types.ContainerListOptions{})
+		types.ContainerListOptions{
+			All: client.checkAll})
```
stffabi, line 74: By setting `All` we would also include `restarting` and `removing` containers, which might have some undefined behaviour if watchtower tries to update them. It would be more robust to add a filter with specific states: `running`, `exited` .

container/container.go
```diff
@@ -27,7 +27,8 @@ func NewContainer(containerInfo *types.ContainerJSON, imageInfo *types.ImageInsp
 
 // Container represents a running Docker container.
 type Container struct {
-	Stale bool
+	Stale      bool
+	WasRunning bool
```
stffabi, line 31: It would be possible to just use the `IsRunning()` function instead of adding this attribute, because the state is determined during the list call and it's immutable for the container instance. It could only change by retrieving a new updated instance from list containers.

main.go
```diff
@@ -61,6 +61,11 @@ func main() {
 			Usage:  "the cron expression which defines when to update",
 			EnvVar: "WATCHTOWER_SCHEDULE",
 		},
+		cli.BoolFlag{
+			Name:   "check-all",
```
stffabi, line 65: What would you think about naming the flag `--include-stopped`?

stffabi (changes requested):
Thanks for your contribution. LGTM, I've added some minor comments.

Would it be possible for you to take look into those?

simonaronsson:
#76 also wishes this feature.

simonaronsson:
This PR is dead and there are still some design issues thar prevents it from being merged.
I'll push an alternative PR solving the issue based on masters current HEAD shortly.

PR #159, opened 2018-01-27
Gitmask Anonymous PR

gitmask-anonymous:
This is an anonymous PR submitted via Gitmask - https://www.gitmask.com


stffabi:
Thanks for your PR. I would prefer not to include additional binaries and glibc into the container. But I changed how the ca-certificates get's included. As of b15df25fc05fdc4896e425750fa1512e8efb018d the ca-certificates will be copied from latest alpine, that should make it more clear where it comes from.

Thanks again for your interest in making watchtower better.

PR #165, opened 2018-02-16
Add discord notification service

darknode:
Send notifications into Discord channel with webhook url.
Part from README.md:

## Notifications
....
* `discord` to send notifications into Discord channel

### Discord
To send notifications into Discord channel, the following command-line options, or their corresponding environment variables, can be set:

* `--notification-discord-webhook` (env. `WATCHTOWER_NOTIFICATION_DISCORD_WEBHOOK`): Webhook URL configured in Discord.

[Checkout how to add webhook for channel](https://support.discordapp.com/hc/en-us/articles/228383668)

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -199,3 +199,8 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+Watchtower contains code that is licensed under a BSD-license:
+ - Copyright (c) 2009 The Go Authors. All rights reserved.
+   
+   For details see https://golang.org/LICENSE
--- a/README.md
+++ b/README.md
@@ -189,7 +189,9 @@ Watchtower can send notifications when containers are updated. Notifications are
 The types of notifications to send are passed via the comma-separated option `--notifications` (or corresponding environment variable `WATCHTOWER_NOTIFICATIONS`), which has the following valid values:
 
 * `email` to send notifications via e-mail
+* `discord` to send notifications into Discord channel
 
+### Email
 To receive notifications by email, the following command-line options, or their corresponding environment variables, can be set:
 
 * `--notification-email-from` (env. `WATCHTOWER_NOTIFICATION_EMAIL_FROM`): The e-mail address from which notifications will be sent.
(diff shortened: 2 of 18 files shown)


stffabi:
Thanks for your contribution, LGTM.

I have only one thing we should discuss. If we want to go the `StartNotification` and `SendNotification` way or if we should just send one message after one. I haven't used discord so far, but as from reading the homepage it seems to me more like something like slack. The current slack implementation sends one message after the other and doesn't collect them.
So it might be good if both behave in a similar way, what do you @darknode think?

darknode:
@stffabi thanks for your review. I agree that chat-like system notifications better implement as you mentioned. I will adjust discord to slack behavior in 2 days. Would you mind if I add MSTeams notification in another pull-request?

stffabi:
Great, thanks for your initiative and adding MSTeams would be awesome, 👍 .

Please make sure to rebase onto master, because of merging of the slack notification system, there are now some minor conflicts.

Racle:
For webhooks, slack based webhooks should work just fine (we use few slack based webhooks on our discord channel).

#113 should work just fine with Discord also :) 

EDIT: I have not tested this feature yet, so I'm not 100% sure that it works flawlessly. Just another slack webhooks work just fine

stffabi:
@Racle thanks for pointing out on that. @darknode then we could just adjust the documentation of slack and point out that it could also be used for discord. What do you think?

stffabi:
@darknode I'm going to release a new version with slack today, maybe you could give it a try and test if it also works on discord?

darknode:
@stffabi I'll try to test it today and share results here.

darknode:
@stffabi @Racle I've checked discord docs and found [a way to use slack payloads with discord](https://discordapp.com/developers/docs/resources/webhook#execute-slackcompatible-webhook). So I close this pull-request and see you in MSTeams notificator PR :)

stffabi:
Great to hear that and thanks for your work.

Racle:
@stffabi to use discord, just use discrod webhook link and add /slack after that.
Just tested slack webhook with discord and it works like it should :)

For improvement, I would like to get only one message when container is updated, just small message which says "Container X has been updated". There is 4 message flood anytime container updates.

PR #208, opened 2018-07-16
Adjust Documentation 

connormcmk:
This pull request attempts to do three things:
1. Fix an errant 'below' that should have been 'above' and causing me confusion
2. Give an example of how to pass environment variables (something not everyone might know) with additional links to resources
3. Add documentation to ensure users with ARM based architecture use the right image. 

My further question is: does mounting the host's docker config file only work with docker compose/swarm mode? If so that should be noted, because I'm only using containers and it caused me no end of trouble trying to get that to work (it never did).

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Every few minutes watchtower will pull the latest *centurylink/wetty-cli* image 
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


simonaronsson:
Thanks for the PR!

PR #265, opened 2019-04-04
Merge changes kopfkrieg/watchtower

KopfKrieg:
As promised a pull request. I'm currently looking through the changes and the only thing I'd like to keep is the new Dockerfile. Everything else are just small changes with no value. Maybe we also should apply gofmt to the codebase once, it fixes a few indentation errors.

Any opinions about it?


simonaronsson:
Merging it to branch kopfkrieg-master so that I may reset the changes we dont want to merge.

simonaronsson:
I'll create a merge request from kopfkrieg-master to master and add you to it, so we may continue discussions there.

PR #268, opened 2019-04-04
Kopfkrieg master

simonaronsson:


simonaronsson:
Actually, I just redid the whole circleci workflow into something a bit more manageable, so this wont even be needed. 👍

Issue #274, opened 2019-04-10
Multi Arch Builds


thelamer:
I have the logic layed out in travis-ci for building and pushing armhf,arm64, and x86_64 builds of watchtower with a manifest so users can just use the :latest tag across different hardware. 

https://github.com/Taisun-Docker/updater/blob/master/.travis.yml
https://github.com/Taisun-Docker/updater/blob/master/Dockerfile.amd64
https://github.com/Taisun-Docker/updater/blob/master/Dockerfile.armhf
https://github.com/Taisun-Docker/updater/blob/master/Dockerfile.aarch64

This was all out of necessity to provide Taisun and Linuxserver.io users an easy to use command to upgrade their containers without having a watch concept. 

I do not know circle well, but it does not look like you are currently using them for your CD pipeline just CI. 

At a minimum if armhf,arm64, and amd64 manifested dockerhub endpoints are available we can swap to recommending our users run the `taisun/updater` method as it is a static fork and start using watchtower with the --run-once flag. 

Given the ease of cross compilation it might also be prudent to add other arches to this manifest list (powerpc etc).

simonaronsson:
Hey!

Thanks for opening a ticket. Im not sure I understand what you are asking here, so feel free to elaborate in case I got it wrong. :D

Right now, we’re building arm64v8, armhf, amd64 and i386 images. This is done both for CI and CD. However, only pushed to dockerhub during CD.

Looking at our current available images I see the arm alternatives as well, though not taggad with a -latest suffix. Is that the part you’re missing? Pulling for exempel containrrr/watchtower:0.3.2 should in other words already work.

🙌🏻

thelamer:
Ahh this new endpoint does not even show up in searches yet. I was basing off the v2tech that has not been updated in a year.

thelamer:
@simskij Re-opening this issue as it boils down to leveraging Docker Manifests, you have tags available for armhf and arm64v8 but when you pull latest or 0.3.1 you get the x86 variant. 

Check out the difference of these commands: 

```
docker manifest inspect taisun/updater:latest
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 948,
         "digest": "sha256:006c11a486ba0829498bb1ef7cb00547ad16c6034a3fa93b3015da37b999668c",
         "platform": {
[...]

PR #278, opened 2019-04-14
#274 additional release logic to try to push manifested releases on publish

thelamer:
I made a lot of assumption here, realistically If is was you I would have identical logic for commits to master as release tags with publishing workflow. 

You can just push to an alternative endpoint like containrrr-dev/watchtower using the git commit as the tag. 
The combination of bash commands and goreleaser is a little confusing here, and I still do not understand how the publishing works fully as you have 386 defined in your goreleaser , but there is no tag for it on dockerhub currently. Also there is no logic for the latest tag as far as I can see.  

Edit: forgot to mention noticed the blank description for your dockerhub repo, At linuxserver we use a simple node script to keep repo READMEs in sync with dockerhub, totally up to you to keep something like that in there. Reference issue https://github.com/containrrr/watchtower/issues/274

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,3 +114,37 @@ jobs:
       - run: 
           name: Execute goreleaser
           command: CGO_ENABLED=${CGO_ENABLED:-0} ../goreleaser --debug
+      - run:
+          name: Docker Manifest
+          command: |
+            mkdir -p ~/.docker/ && \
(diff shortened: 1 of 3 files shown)


simonaronsson:
Thanks a bunch, I'm gonna tweak it a bit and take it for a test drive. Hope you dont mind! 😄

PR #282, opened 2019-04-15
check if schedule len > 0 to prevent collisions

### LINKED ISSUE #275: Fatal error about both interval and schedule being set, even though only interval is set
**Describe the bug**
Unable to start watchtower due to error regarding both interval and schedule being set, however only interval is set.

**To Reproduce**
Compile watchtower based on latest commit in master, and run `watchtower --debug --cleanup --host=unix:///var/run/docker.sock --interval=300` this fails with error below. As you can see only interval is set, schedule is not.

**Expected behavior**
Every 5 mins watchtower would check to see if anything needed updates.

**Environment**
 - Platform Digital Ocean Virtual Machine (OS: Ubuntu)
 - Architecture x86_64
 - Docker version 18.03.0-ce

**Logs from running watchtower with the `--debug` option**
time="2019-04-12T03:03:14Z" level=fatal msg="Only schedule or interval can be defined, not both."

techknowlogick:
Fix #275

### DIFF
--- a/main.go
+++ b/main.go
@@ -66,10 +66,11 @@ func before(c *cli.Context) error {
 
 	pollingSet := c.IsSet("interval")
 	cronSet := c.IsSet("schedule")
+	cronLen := len(c.String("schedule"))
 
-	if pollingSet && cronSet {
+	if pollingSet && cronSet && cronLen > 0 {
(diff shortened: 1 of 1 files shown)


simonaronsson:
Thanks a bunch for the PR! Merging!

Issue #281, opened 2019-04-15
0.3.3 appears to break email notifications


SmallFriendlyKiwi:
With version 0.3.3, I no longer get email notifications.

All versions of containrrr/watchtower up to 0.3.2 send me email notifications, as desired.
v2tec/watchtower sends emails notifications, as desired

**To Reproduce**
This is my Dockerfile that works and sends emails (0.3.2):

docker run -d \
  --name watchtower \
  --hostname=azure.dockerhost001 \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -e WATCHTOWER_NOTIFICATIONS=email \
  -e WATCHTOWER_NOTIFICATION_EMAIL_FROM=admin@itinnovations.co.nz \
  -e WATCHTOWER_NOTIFICATION_EMAIL_TO=admin@itinnovations.co.nz \
  -e WATCHTOWER_NOTIFICATION_EMAIL_SERVER=email-smtp.us-east-1.amazonaws.com \
  -e WATCHTOWER_NOTIFICATION_EMAIL_SERVER_USER=REMOVED \
  -e WATCHTOWER_NOTIFICATION_EMAIL_SERVER_PASSWORD=REMOVED \
  -- containrrr/watchtower:0.3.2 \
  --cleanup

This is my Dockerfile that works but does not send emails (0.3.3)

docker run -d \
  --name watchtower \
  --hostname=azure.dockerhost001 \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -e WATCHTOWER_NOTIFICATIONS=email \
  -e WATCHTOWER_NOTIFICATION_EMAIL_FROM=admin@itinnovations.co.nz \
  -e WATCHTOWER_NOTIFICATION_EMAIL_TO=admin@itinnovations.co.nz \
  -e WATCHTOWER_NOTIFICATION_EMAIL_SERVER=email-smtp.us-east-1.amazonaws.com \
  -e WATCHTOWER_NOTIFICATION_EMAIL_SERVER_USER=REMOVED \
  -e WATCHTOWER_NOTIFICATION_EMAIL_SERVER_PASSWORD=REMOVED \
  -- containrrr/watchtower:0.3.3 \
  --cleanup

**Expected behavior**
Email notifications should continue to be received in 0.3.3

**Environment**
 - Platform:
    Azure VM

 - Architecture
PRETTY_NAME="Debian GNU/Linux 8 (jessie)"
NAME="Debian GNU/Linux"
[...]

PR #283, opened 2019-04-19
Add Dockerfile.self-contained

svengo:
This pull request adds an alternative dockerfile that uses docker's multistage build feature to build watchtower in a container and not on the host system. 

I've been using a very similar docker file for a while to build [svengo/watchtower](https://hub.docker.com/r/svengo/watchtower). Maybe someone else will find that useful, too. See also [#281](https://github.com/containrrr/watchtower/issues/281#issuecomment-484467347).

### DIFF
--- a/Dockerfile.multistage-build
+++ b/Dockerfile.multistage-build
@@ -1,0 +1,39 @@
+#
+# Builder
+#
+
+FROM golang:alpine as builder
+
+# use version (for example "v0.3.3") or "master"
+ARG WATCHTOWER_VERSION=master
+
+RUN \
+  apk add --no-cache \
+    alpine-sdk \
+    ca-certificates \
+    git \
+    tzdata && \
+  \
+  mkdir --parents $GOPATH/src/github.com/containrrr && \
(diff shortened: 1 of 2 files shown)


simonaronsson (changes requested):
Thanks for your contribution. It's very appreciated!

Would you mind renaming the file Dockerfile.self-contained as both the CI Dockerfile and this one uses multi-stage builds.

Also, are all the command chaining really needed? I think the following changes would improve it further by making it easier to read:

* one RUN for all the apk adds
* no need to put it in the gopath as we're using go modules, just clone it where you stand 😜 
* put the go build and go test in their own RUN

Does this make sense?
Thanks again 🙏

svengo:
Thanks for the review and the suggestions, @simskij!

I have implemented the changes in 1a9f217. Normally I try to minimize the number of RUN commands and layers, but since the build-image is thrown away anyway, I don't see any disadvantages. The naming makes sense.

simonaronsson (approved):

PR #289, opened 2019-04-28
Make it possible to use watchtower to update exited or created containers as well

### LINKED ISSUE #112: Updating images for stopped containers

simonaronsson:
Resolves #112 

Adds the flag `--include-stopped` which, when used, makes watchtower pull images and recreate exited and created containers as well. This feature defaults to disabled. Appropriate tests has been added to `container_test.go`.

Would appreciate if someone other than me tried it out before merging it with master. 🙏 
@KopfKrieg and/or @Codelica, any of you up for the task? 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ docker run --rm containrrr/watchtower --help
 - `--tlsverify` Use TLS when connecting to the Docker socket and verify the server's certificate.
 - `--debug` Enable debug mode. When this option is specified you'll see more verbose logging in the watchtower log file.
 - `--monitor-only` Will only monitor for new images, not update the containers.
+- `--include-stopped` Will also include created and exited containers.
 - `--help` Show documentation about the supported flags.
 
 See below for options used to configure notifications.
--- a/app/app.go
+++ b/app/app.go
@@ -155,8 +155,14 @@ func SetupCliFlags(app *cli.App) {
 			EnvVar: "WATCHTOWER_MONITOR_ONLY",
 		},
 		cli.BoolFlag{
-			Name:  "run-once",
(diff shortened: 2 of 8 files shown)


Codelica:
Interesting...  I guess this could potentially resurrect a container if a bad image was deployed.  I would be interested in trying it out, but to be honest I won't be until the end of the week when I can even look.  My load is just too heavy at the moment.  :/

simonaronsson:
too bad no one was able to aid me in reviewing this change. to be able to proceed i've done some additional manual testing as well as running the unit test suite multiple times. this leads me to the conclusion that no apparent critical bugs have been introduced. with that said; i'll go ahead and merge this. if it introduces any anomalies or duplication behavior, please let me/us know immediately.

simskij:
scratch that, i've actually found a bug 😆

Codelica:
Sorry about that..  spent the whole weekend at a coding competition for my daughter.  Between that and projects at work, there's been little "fun" time the last couple weeks.  That should change once my work project clears ;)

zoispag:
> Interesting... I guess this could potentially resurrect a container if a bad image was deployed.

Hi @simskij and sorry for replying in an old "thread", but I tried to avoid making a new issue unless necessary. The `--include-stopped` works like a charm and will pull new/delete old image for stopped containers as well. But from the PR changes and the tested behavior, I see that it doesn't start the container again. Is it possible somehow to do so? I don't see an argument, unless I am missing something, or maybe there is a way with the pre-check. Not sure. Please advice :)

PR #328, opened 2019-05-25
Update option description

simonaronsson:
> I added order alphabetically the options, adding information like Environment Variable. I took the informations from app.go file.

@all-contributors add @ansem93 for doc


simonaronsson:
@all-contributors add @Ansem93 for doc

PR #340, opened 2019-06-17
Initial work for the URL callback

grazzolini:
Adding configuration for the URL callback


simonaronsson:
Whats this? :)

grazzolini:
@simskij This is something I've been working on for allowing watchtower to use an out of band server for knowing which images to update. But, since this is not yet complete, I've closed the PR. I plan on submitting a PR later when we have the full implementation done.

PR #373, opened 2019-08-25
add pre/post update check lifecycle hooks

rymdo:
Added more lifecycle hooks

### DIFF
--- a/internal/actions/update.go
+++ b/internal/actions/update.go
@@ -13,6 +13,8 @@ import (
 func Update(client container.Client, params UpdateParams) error {
 	log.Debug("Checking containers for updated images")
 
+	executePreCheck(client, params)
+
 	containers, err := client.ListContainers(params.Filter)
 	if err != nil {
 		return err
@@ -36,12 +38,14 @@ func Update(client container.Client, params UpdateParams) error {
 	checkDependencies(containers)
 
 	if params.MonitorOnly {
+		executePostCheck(client, params)
 		return nil
 	}
 
 	stopContainersInReversedOrder(containers, client, params)
 	restartContainersInSortedOrder(containers, client, params)
 
+	executePostCheck(client, params)
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


internal/actions/update.go
```diff
@@ -36,12 +38,14 @@ func Update(client container.Client, params UpdateParams) error {
 	checkDependencies(containers)
 
 	if params.MonitorOnly {
+		executePostCheck(client, params)
```
simonaronsson, line 41: This is a bit weird, isn't it? monitor only is supposed to not run anything that might mutate the containers.
Both `executePreCheck` and `executePostCheck` should be skipped if the user runs watchtower with `--monitor-only` enabled in my opinion.

internal/actions/update.go
```diff
@@ -13,6 +13,8 @@ import (
 func Update(client container.Client, params UpdateParams) error {
 	log.Debug("Checking containers for updated images")
 
+	executePreCheck(client, params)
```
simonaronsson, line 16: See below

internal/actions/update.go
```diff
@@ -123,11 +127,58 @@ func checkDependencies(containers []container.Container) {
...
+	containers, err := client.ListContainers(params.Filter)
+	if err != nil {
+		return
+	}
+	for _, container := range containers {
+		executePostCheckCommand(client, container)
+	}
+}
+
+func executePreCheckCommand(client container.Client, container container.Container) {
```
simonaronsson, line 150: Should we maybe extract this logic to a shared method doing all the work, taking the step and the command as a argument? Ie. `executePostCheck(client container.Client, params UpdateParams)` calling `executeCommandInContainer(client container.Client, params UpdateParams, command string, step string)? This includes `executePreUpdate` and `executePostUpdate` as well.

simonaronsson (changes requested):
Thank you for the awesome contribution! 🙏

All in all, this looks good to me. Added some minor pointers. I'd also like to request that you update the documentation to show how to use the newly added feature.

internal/actions/update.go
rymdo: I get your point. One could also see them as independent of each other as you need to activate hooks explicitly. Hooks also don't necessarily have to mutate any container and it could be unnecessarily limiting just to block them for that reason. If we change so that the hooks is skipped on "--monitor-only" I suggest that a warning of some kind is thrown when combined with "--enable-lifecycle-hooks".

simonaronsson (approved):
LGTM

Issue #264, opened 2019-04-02
Additional notification options


simonaronsson:
Moving this discussion to a [issue in shoutrrr](https://github.com/containrrr/shoutrrr/issues/2).

Thanks for understanding! 🙏

Codelica:
I like it, just wish Pushover were higher.  😇 I can help test that and Webhooks when the time comes.

Also for Webhooks please consider header support if possible, for Bearer authentication.

simonaronsson:
Lets move it up a bit then :)

simonaronsson:
These will be solved by replacing the current notification implementation with another project I'm working on; [containrrr/shoutrrrr](https://github.com/containrrr/shoutrrr). Whether or not this will be in 0.4.0 or in a later release is not yet written in stone and depends a lot on whether im doing it by myself or with help. 😄

simonaronsson:
Discord, telegram, slack and pushover are all implemented in shoutrrr as of [5314ee954ba8dbc1c6eb6cbde310af695c8ae1da](https://github.com/containrrr/shoutrrr/commit/5314ee954ba8dbc1c6eb6cbde310af695c8ae1da). Will implement a couple more before i'll even attempt integrating it.

Codelica:
FWIW... I'm no Go expert, but I did made a little test run using a Pushover URL in Shoutrrr and it worked fine.  :)

somq:
Just to say, HTTP webhook is a killer feature imo.

MAGICCC:
Neat, so we should wait until the shoutrrr code is in watchtower to use Telegram notifications?

simonaronsson:
Yeah. Feel free to help out though if you want it to happen sooner rather than later! :-)

alexanderadam:
Just to be sure: would it be possible to integrate more than one service with [shoutrrr](https://github.com/containrrr/shoutrrr#as-a-package)?
So i.e. to have notifications via email, XMPP and Sentry **at the same time**?

simonaronsson:
> Just to be sure: would it be possible to integrate more than one service with [shoutrrr](https://github.com/containrrr/shoutrrr#as-a-package)?
> So i.e. to have notifications via email, XMPP and Sentry **at the same time**?

yeah, that's our thought exactly! 👍 we've not really decided on how to do this in the case of watchtower, but yes - that will be possible in a general.

noplanman:
[...]

PR #379, opened 2019-09-09
Add support for Hangouts Chat notifications

erseco:
This adds support for Hangouts Chat notifications. 

https://developers.google.com/hangouts/chat/how-tos/webhooks

Work items:

A flag has been added to internal/flags/flags.go:
"notification-hangouts-url".

A Hangouts Chat notification driver has been added in
notifications/hangouts.go.

"hangouts" has been added to notification driver choices in
notifications/notifier.go.

Docs have been updated


alexanderadam:
I'm not quite sure how likely it is that this is gonna be merged because AFAIK [all notification protocols will be moved from watchtower into shoutrrr](https://github.com/containrrr/watchtower/issues/264#issuecomment-485149945).
So it might be better to make a PR [for Shoutrrr](https://github.com/containrrr/shoutrrr) instead.

**Disclaimer:** I'm not a maintainer, so I have no idea whether it will or whether it won't be merged.

simonaronsson:
As @alexanderadam states, our current goal is to switch out all notification mechanics from watchtower with shoutrrr. I'd rather not merge this as it would stop working again as soon as that work is started. Feel free to provide a PR for shoutrrr though. I myself or @piksel will assist in getting it added.

PR #384, opened 2019-10-02
Fix a small typo

foosel:
Just something I noticed while playing around.


simonaronsson:
Thanks for the contribution!

simonaronsson:
@all-contributors add @foosel for code

Issue #388, opened 2019-10-09
Add mail customized mail notification subject tag


8ear:
Hi all,

Your Watchtower container is very nice and the mail function works as excepted. Really great and a huge thanks to you!

I have more than one Docker server with different purposes and on every server I have a watchtower container running in monitoring mode. But the mails I got have all the following subject: `Watchtower updates on <container id>` 
For example: `Watchtower updates on 1a25cbf5f088`.

So my Question is, is it possible to add an environment variable to add an tag in the subject line at the beginning which can be customized?

The result can be:
`[TAG] Watchtower updates on <container id>` for example: `[server 13] Watchtower updates on 1a25cbf5f088`.
ENV: WATCHTOWER_NOTIFICATION_EMAIL_SUBJECTTAG=[server 13]

Kind regards

simonaronsson:
Thank you for taking care of this. I'll go ahead and switch the status of this ticket to `In Progress` meanwhile.

8ear:
Thanks.

simonaronsson:
This is already merged. Closing.

PR #389, opened 2019-10-09
Added Mail Subject Tag to email.go

8ear:
Hi,
I am not an Go developer. Therefore please review my code. :-)

I would add an customizable subject tag for mail notification as I described in the issue #388 .

Thanks to you!

PS: I can also change the documentation if this code is ok.

### DIFF
--- a/pkg/notifications/email.go
+++ b/pkg/notifications/email.go
@@ -24,7 +24,7 @@ const (
 // We work around that by holding on to log entries until the update cycle is done.
 type emailTypeNotifier struct {
 	From, To               string
-	Server, User, Password string
+	Server, User, Password, SubjectTag string
 	Port                   int
 	tlsSkipVerify          bool
 	entries                []*log.Entry
@@ -43,6 +43,7 @@ func newEmailNotifier(c *cobra.Command, acceptedLogLevels []log.Level) t.Notifie
 	port, _ := flags.GetInt("notification-email-server-port")
 	tlsSkipVerify, _ := flags.GetBool("notification-email-server-tls-skip-verify")
 	delay, _ := flags.GetInt("notification-email-delay")
+	subjecttag, _ := flags.GetInt("notification-email-subjecttag")
 
 	n := &emailTypeNotifier{
 		From:          from,
@@ -54,6 +55,7 @@ func newEmailNotifier(c *cobra.Command, acceptedLogLevels []log.Level) t.Notifie
 		tlsSkipVerify: tlsSkipVerify,
 		logLevels:     acceptedLogLevels,
 		delay:         time.Duration(delay) * time.Second,
+		SubjectTag:    subjecttag,
 	}
 
 	log.AddHook(n)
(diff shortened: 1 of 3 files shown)


pkg/notifications/email.go
```diff
@@ -62,7 +64,14 @@ func newEmailNotifier(c *cobra.Command, acceptedLogLevels []log.Level) t.Notifie
 }
 
 func (e *emailTypeNotifier) buildMessage(entries []*log.Entry) []byte {
-	emailSubject := "Watchtower updates"
+	// If variable SubjectTag is empty define emailSubject as "Watchtower updates".
+	// If variable SubjectTag is set define emailSubject as "$SubjectTag Watchtower updates"
+	// For example: SubjectTag=[Server Munich] -> "[Server Munich] Watchtower updates ..."
+	if SubjectTag == nil {
```
simonaronsson: a string is never nil, however it could be "" if left unset. however, if you go for the suggestion with replacing the whole subject tag, no comparison is needed at all. Also, I'd prefer if you dont leave comments in the code explaining something that is perfectly understandable from reading the code itself 👍  😄

pkg/notifications/email.go
```diff
@@ -43,6 +43,7 @@ func newEmailNotifier(c *cobra.Command, acceptedLogLevels []log.Level) t.Notifie
 	port, _ := flags.GetInt("notification-email-server-port")
 	tlsSkipVerify, _ := flags.GetBool("notification-email-server-tls-skip-verify")
 	delay, _ := flags.GetInt("notification-email-delay")
+	subjecttag, _ := flags.GetString("notification-email-subjecttag")
```
simonaronsson, line 46: The flag also needs to be added to internal/flags/flags.go, I'd actually prefer if it had a overridable default value, like the previous "Watchtower Updates" and replaced the subject altogether.

simonaronsson (changes requested):
Thank you so much for your contribution! I've left a two-part minor change request that I'd like you to have a look at before this will be ready to merge. 🙏

simonaronsson:
Also, do note that the code might be changed when we integrate shoutrrr. But until then, whenever that is, this is a great minor addition. 😅

pkg/notifications/email.go
8ear: Ok. Thx. I will change it or can you change it

8ear:
I understand, when it is planned to integrate shoutrrr?

pkg/notifications/email.go
8ear: Ok. I will think about it.

simonaronsson:
I dont know at this point to be honest. It's ready to be integrated (ish) but I've just not found the time to actually do it.

simonaronsson (approved):

simonaronsson (changes requested):
Oh, one other small thing! Would you mind updating the documentation as well? There are previous examples on how to document arguments/flags available. After that we're good to go from my perspective! 👍

simonaronsson:
Fixed the documentation. Merging. Thanks for your contribution! 🙏🏼

simonaronsson:
@all-contributors add @8ear for code

PR #401, opened 2019-11-06
Add lifecycle hooks to documentation nav

zoispag:
Add lifecycle hooks to documentation nav, because now it can be found only via search.


simonaronsson:
Awesome! Thanks for your contribution! I must have missed this 😅

PR #418, opened 2019-12-18
Don’t delete cmd when runtime entrypoint is different

jsclayton:
This fixes an issue where you've overridden and _changed_ the `entrypoint`, overridden `cmd` with identical params, and when watchtower updates it the `cmd` is lost on the newly created container. 

Context:

You have a `Dockerfile` with the following:

```
CMD ["node", "index.js"]
ENTRYPOINT ["/app/entrypoint-1.sh"]
```

Suppose that in another context you want to use a different entrypoint. The issue is that, in both [Docker Compose](https://docs.docker.com/compose/compose-file/#entrypoint) and [Docker run](https://docs.docker.com/engine/reference/run/#entrypoint-default-command-to-execute-at-runtime), if you provide an alternate entrypoint, it will clear out any command set in the Dockerfile. 

When watchtower is diffing the running container and the image it was started from, it's considering identical values to be defaults from the image. When the entrypoint is the same, this is fine and works as expected. When the entrypoint is different, but you've defined the same start command for the container per the above warnings, the command is considered default and not included in the updated container. 

This PR intends on fixing this by considering identical commands to be defaults only when the entrypoint is also identical. 

I'll add that while I _think_ this is the fix for the problem, I haven't been able to find a good way to test or confirm it short of continuously pushing changes to a registry. If there's a way to test it locally, or force container recreation without an update of the source image, I'd be happy to test it.


simonaronsson:
One thing I usually do when I test is to force an update.
For example, if you create a portainer container:
```zsh
$ docker tag tutum/hello-world portainer/portainer:latest
$ docker run -d --name portainer-test portainer/portainer
$ docker logs --follow watchtower
```

Watchtower should then, during the next update, see that the hash differs and update it to the image currently set as :latest.

simonaronsson:
Oh, while the change looks good to me, I'd really appreciate it if you would consider testing it using the method described above before I merge the change.

jsclayton:
@simskij Thanks for the test example, **much** simpler than I had been trying pushing to a registry. 😌 

Looks like it's working fine. The output shows the upgrade to the actual portainer image:

```
➜  watchtower git:(fix/retain-cmd) ./watchtower -d -i 30
DEBU[0000] Sleeping for a second to ensure the docker api client has been properly initialized.
DEBU[0001] Retrieving running containers
DEBU[0001] There are no additional watchtower containers
DEBU[0001] Starting Watchtower and scheduling first run: 2020-02-26 17:07:21 -0700 MST m=+31.057211219
DEBU[0031] Checking containers for updated images
DEBU[0031] Retrieving running containers
DEBU[0031] No pre-check command supplied. Skipping
DEBU[0031] Retrieving running containers
DEBU[0031] Pulling portainer/portainer:latest for /portainer-test
DEBU[0031] No credentials for portainer in /config.json
DEBU[0031] Got auth value:
DEBU[0031] Got image name: portainer/portainer:latest
DEBU[0031] No authentication credentials found for portainer/portainer:latest
INFO[0037] Found new portainer/portainer:latest image (sha256:10383f5b5720d7e1f5f824137034c69b7f6d82cc8aa33afcc4e9d508b561af77)
DEBU[0037] No pre-update command supplied. Skipping
INFO[0037] Stopping /portainer-test (99978c762268ee5be26deb4618e057395647502a5811c62878e31cd87a56d89d) with SIGTERM
DEBU[0038] Removing container 99978c762268ee5be26deb4618e057395647502a5811c62878e31cd87a56d89d
INFO[0038] Creating /portainer-test
DEBU[0038] Starting container /portainer-test (3b08c41429e09afd97ad31c4ea9b9348ad8500489b6b36d7ad5ff6902eca2d9c)
DEBU[0039] Retrieving running containers
DEBU[0039] No post-check command supplied. Skipping
DEBU[0039] Scheduled next run: 2020-02-26 17:07:51 -0700 MST
```

The container's image hash also changed after the update.

simonaronsson:
@all-contributors add @jsclayton for code

simonaronsson:
Thank you for your contribution! Much appreciated!

PR #424, opened 2019-12-25
Added section about private registries to documentation

jnidzwetzki:
I needed some time to get watchtower running with a private image registry. In my opinion this point is missing in the documentation so far. To make it easier for other users, I would like to add my experiences to the official documentation.


simonaronsson:
Thank you very much for your excellent contribution! 🙏🏼

Issue #430, opened 2020-01-07
Self contained docker build for local changes


victorcmoura:
While building Watchtower containers with the default Dockerfile, I bumped into a few errors related to how my go build was being created. I tried using the self-contained Dockerfile and it worked. 

There should be a development version of the self-contained build to enable contributors to safely build docker images of their own versions of Watchtower without having to worry about environment configs.

simonaronsson:
Solved through #431

PR #431, opened 2020-01-07
Adds self-contained dev Dockerfile

victorcmoura:
This PR solves #430 by adding a custom self-contained Dockerfile which, instead of cloning the repository and building the image from it, it copies the local watchtower folder containing the developer's changes.


victorcmoura:
Since there would be 3 Dockerfiles in the root folder, I found it better to put them in a separate one: 503d693

zoispag (changes requested):
Moving the main Dockerfile will break the build/release process.

zoispag:
Thanks!

victorcmoura:
@zoispag Thanks for the heads up. I had already made the change but forgot to add it to the commit =P

zoispag:
@victorcmoura I suggest to add a reference in CONTRIBUTING.md file, related to how to use this Dockerfile.
I will let Simon to review further.. 😎

victorcmoura:
@zoispag It seemed pretty fair to add additional instructions in CONTRIBUTING.md. Done =)

zoispag (approved):

simonaronsson:
Thank you so much for your contribution! 🙏

To be totally honest though, I've always had a hard time finding a use case for the self-contained dockerfile. What is the benefit of this compared to running `goreleaser --snapshot --skip-publish`?

I don't mean to bash or anything, just want to know whether I'm missing something or if the docs just need to be clearer on how to build local dev images using goreleaser?

victorcmoura:
Hi @simskij. 

Given my personal experience, I believe that if the docs were clearer about generating it with goreleaser, I would've gone for it. CONTRIBUTING.md only showed how to build the binary itself. I tried a simple `docker build .` with the default Dockerfile but I bumped into some errors on `docker run`, even after following the commands CONTRIBUTING.md was listing. Whereas, with the self-contained build, I managed to do it gracefully.

And, as @zoispag suggested, I added the instructions, in CONTRIBUTING.md, on how to use them to build an image. I believe it is more intuitive for newcomers (either in Watchtower or Golang), to go for a `docker ...` command than for a `goreleaser ...` when they're attempting to build an image of their own.

PS: This is an opinion from a newcomer in Golang that had never heard of goreleaser haha

simonaronsson:
Thanks for the clarifiation, @victorcmoura! I'll let this pass for now, although I believe we need to have a discussion at some point regarding how we want to build, debug and release watchtower. Multiple ways of achieving the same thing are likely to become an issue down the line.

Issue #429, opened 2020-01-07
Watchtower HTTP API proposal


victorcmoura:
Since Watchtower basically watches for docker registries and actively pulls images in order to check for outdated containers, it keeps continuously incrementing the registries' pull counter, making them useless. The counts can no longer be considered as user downloads/pulls.

This issue has been impacting some of my co-workers' personal projects and I have managed to use it as a working case in my bachelor's thesis. My proposal is to implement an HTTP API to enable DockerHub's Webook service to trigger Watchtower instances and make them check for updates only when a new image is pulled. Another use case would be a request made by a CI platform in order to trigger the update process.

I have already started some prototypes and I hope to be implementing a working version soon.

dinuta:
Hi @victorcmoura 

Can you also consider update fallback in the api implementation? 

https://github.com/containrrr/watchtower/issues/427

victorcmoura:
Hi, @dinuta. As I mentioned in the PR, I believe it is safer not to overwhelm, with more complex features, the first implementation and keep them for future contributions. But it is certainly something that I see as necessary to make Watchtower usage more feasible in production environments (at least small production environments).

victorcmoura:
Closed in https://github.com/containrrr/watchtower/pull/432

PR #432, opened 2020-01-08
Watchtower HTTP API based updates

victorcmoura:
This pull request implements the changes proposed in #429.

API Features + ToDo List:

- [x] **/v1/update** endpoint to trigger image updating process
- [x] **--http-api** flag to set up HTTP API mode
- [x] ~**--http-api-port** flag to customize the listening port~ (Container will always listen to 8080)
- [x] authentication token set by ENV variable
- [x] Expose port 8080 in the Dockerfile
- [x] Docs on how to use the API

Required changes:

- [x] Lock to prevent concurrent updates
- [x] API has a package of its own
- [x] validates if the authentication token has been set, makes fatal exit if not

Suggestions are welcomed!


zoispag (commented):
This is useful. Will also solve an issue I have with a container getting updated, while the UI tests are running!! Great 👍🏼 
I think it is a good idea to Expose port 8080 from the Dockerfile, so that an update can be triggered by another running container, without necessarily need to bind a port to the host machine.

zoispag:
Auth should be added.
Maybe a simple header auth, a bearer token probably, that can be provided as an ENV variable to watchtower.

victorcmoura:
@zoispag Just added the suggestions to the feature list. Thank you =)

victorcmoura:
Even though this API presents a huge potential, I believe we've reached a minimal working version that can be improved by future PR's. Let me know if there should be something else implemented by now.

dinuta:
Hi guys, great idea. I just wonder if you can add a rollback endpoint which should do a rollback in case the new image has issues (Yes it's possible )

I think here about 2 possible endpoints.
-  **/v1/rollback** - which will rollback all images under the watchtower instance
-  **/v1/rollback**
Http header: **Rollback-Image** the specific image to rollback.
Example:
Rollback-Image: alpine-latest

docs/http-api-mode.md
```diff
@@ -0,0 +1,35 @@
...
+    labels:
+      - "com.centurylinklabs.watchtower.enable=true"
+
+  watchtower:
+    image: containrrr/watchtower
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+    command: --debug --http-api
+    environment:
+      - WATCHTOWER_HTTP_API_TOKEN=mytoken
```
dinuta, line 24: Hi man,

WATCHTOWER_HTTP_API_TOKEN is too long. Please use HTTP_API_TOKEN or anything you want shorter

simonaronsson: I disagree. This is perfectly fine as we want to promote readability and clarity over brevity.

simonaronsson:
> Even though this API presents a huge potential, I believe we've reached a minimal working version that can be improved by future PR's. Let me know if there should be something else implemented by now.

Thank you, Victor, for your excellent work! I'll have a look as soon as I can.

simonaronsson:
So sorry for allowing this to fall between the cracks. Hope to get to it as soon as possible! Still feels like a great extension of watchtower in my opinion!

simonaronsson:
> Hi guys, great idea. I just wonder if you can add a rollback endpoint which should do a rollback in case the new image has issues (Yes it's possible )
> 
> I think here about 2 possible endpoints.
> 
> * **/v1/rollback** - which will rollback all images under the watchtower instance
> * **/v1/rollback**
>   Http header: **Rollback-Image** the specific image to rollback.
>   Example:
>   Rollback-Image: alpine-latest

Hi!

I know this is a popular request but as we only track by hash on a tag, I really don’t see a feasible way of accomplishing this without having to add some kind of persisted memory to watchtower, which I’m kind of hesitant to do.

If you have any suggestion on how this solution would work, feel free to open an issue for it and submit as much details as possible.

With that said, at least to me, those endpoints are not within the scope of this particular pull request. 

Thanks for understanding! 🙏🏻

cmd/root.go
```diff
@@ -111,6 +114,34 @@ func PreRun(cmd *cobra.Command, args []string) {
...
 	filter := filters.BuildFilter(names, enableLabel)
 	runOnce, _ := c.PersistentFlags().GetBool("run-once")
+	httpAPI, _ := c.PersistentFlags().GetBool("http-api")
+	
+	if httpAPI {
+		log.Println("Watchtower HTTP API started.")
+		
+		apiToken, _ := c.PersistentFlags().GetString("http-api-token")
+
+		http.HandleFunc("/v1/update", func(w http.ResponseWriter, r *http.Request){
```
simonaronsson: I think it would make sense to put this into a separate package, like `pkg/api`. What's your thought on that?

cmd/root.go
```diff
@@ -111,6 +114,34 @@ func PreRun(cmd *cobra.Command, args []string) {
...
+				return
+			}
+
+			if r.Header.Get("Token") != apiToken {
+				log.Println("Invalid token. Not updating.")
+				return
+			}
+
+			log.Println("Valid token found. Triggering updates.")
+			runUpdatesWithNotifications(filter)
```
simonaronsson: This might be a bit dangerous as we'd risk starting an update while another one is running. Maybe we need to add some kind of locking mechanism to this call to make it return something like "Already running an update" in case you try to trigger multiple updates?

cmd/root.go
```diff
@@ -111,6 +114,34 @@ func PreRun(cmd *cobra.Command, args []string) {
 func Run(c *cobra.Command, names []string) {
 	filter := filters.BuildFilter(names, enableLabel)
 	runOnce, _ := c.PersistentFlags().GetBool("run-once")
+	httpAPI, _ := c.PersistentFlags().GetBool("http-api")
```
simonaronsson, line 115: To promote safe practices, I think we should do a fatal exit in case the token is unset/empty.

simonaronsson (changes requested):
Some minor things, but other than that, I'd say this is pretty much good to go. Great work, @victorcmoura! I really appreciate you taking the time to push this change through. 🙏🏼

If you don't feel up for splitting out the API parts to a separate package at this point, just let me know and I'll add it to do the todo list. It's not critical, just something I think would be nice at some point down the line.

cmd/root.go
victorcmoura: I believe it's the best thing to do at the moment. If we let this for future contributions, it may be something that will never be done. I'll work on it.

cmd/root.go
victorcmoura: I've not paid attention to the lock in `runUpgradesOnSchedule`. I'll try to reproduce it in the API package.

cmd/root.go
victorcmoura: I believe you are referring to `apiToken, _ := c.PersistentFlags().GetString("http-api-token")` instead of `httpAPI, _ := c.PersistentFlags().GetBool("http-api")`. And I agree with you. I'll put a conditional to validate it.

victorcmoura:
@simskij, just pushed the requested changes. xD

simonaronsson:
That is awesome! Merging. Thanks again, @victorcmoura!

simonaronsson:
@all-contributors add @victorcmoura for code and docs

simonaronsson (approved):

PR #448, opened 2020-02-07
Unset Hostname when NetworkMode is container

### LINKED ISSUE #188: error msg="Error response from daemon: conflicting options: hostname and the network mode"
I believe this issue happens when containers that are routed through my VPN container (dperson/openvpn) using --net=container:vpn are updated and then restart is attempted but fails due to trying to reuse the old network mode entry which seems to change when the OpenVPN container is restarted also.

Is it possible to set Watchtower to just use a full Docker run command as this is all I have to do to get the container working properly again?

raymondelooff:
The hostname can not be set on containers using a container network stack. Fixes #188.

### DIFF
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -138,6 +138,7 @@ func (c Container) StopSignal() string {
 // the options overridden at runtime.
 func (c Container) runtimeConfig() *dockercontainer.Config {
 	config := c.containerInfo.Config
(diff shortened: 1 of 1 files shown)


simonaronsson:
Thank you so much for your contribution! LGTM

simonaronsson:
@all-contributors add @raymondelooff for code

PR #452, opened 2020-02-21
go.mod: Update golang.org/x/crypto to fix vulnerability

ifraixedes:
Update the golang.org/x/crypto package to fix the vulnerability
CVE-2020-9283.

See https://groups.google.com/forum/#!topic/golang-nuts/XDqhhjZViNk


ifraixedes:
Sorry, I was checking it in a different commit than the lat one.

Issue #456, opened 2020-02-27
Watchtower should additionally set the maximum memory per container


touramakan:
**The maximum memory usage per container should be limitable if this feature is toggle on.**
If the corresponding flag is active watchtower should additionally set the configured limit of maximum memory each container can consume.

**Possible solution**

- Introduce one flag to toggle the feature on/off
- Introduce a second flag to set the desired maxMemory
- Depending on the first flag set the max memory for each container, which does not have a memory less than the second flag

**Alternativ solution**
Start a new project on basis of watchtower (scan containers, stop/start with dependencies consideration) but without the functionality image update 

**Additional context**
As platform provider we face the challenge that end-user deploys their containers (apps), which sometime consume the whole available memory. Since we are already using the watchtower, we would like to extended with this feature.

simonaronsson:
Unfortunately, I don't really think it fits within the scope of the project. I'd recommend that you set up a separate small go binary to do those checks for you, or maintain that feature in a fork.

To me, the use case of watchtower is quite clear, and I'd prefer not to alter that at this point but focus on providing the best experience possible for that.

Thank you for understanding, and hope you find a reasonable way forward. 🙏

PR #457, opened 2020-02-27
Feature/456 Control memory usage of containers

touramakan:
Implementation of the feature request https://github.com/containrrr/watchtower/issues/456


simonaronsson:
Hey! Thnk you for your contribution! I just wanted to let you know that I have not forgot about you, just been short on time. A review is coming as soon as possible :)

touramakan:
> Hey! Thnk you for your contribution! I just wanted to let you know that I have not forgot about you, just been short on time. A review is coming as soon as possible :)

No problem. I´ve just realised, that I need to rework a bit my solution. I´ll let you know, when I am done.

simonaronsson:
I've since had some time to go through your changes, and while I do appreciate your effort **a lot** I'm having a bit of a hard time understanding why the changes are needed.

Isn't this behavior already controlled by the docker flag `--memory`? In that case, I think it would be preferable to avoid duplicating the effort by adding similar control mechanics to watchtower as well. 

What I'm trying to say is; can you elaborate a bit on why this is needed in watchtower as my immediate reaction that it is a bit out of scope?

Thanks 🙏🏼

touramakan:
> I've since had some time to go through your changes, and while I do appreciate your effort **a lot** I'm having a bit of a hard time understanding why the changes are needed.
> 
> Isn't this behavior already controlled by the docker flag `--memory`? In that case, I think it would be preferable to avoid duplicating the effort by adding similar control mechanics to watchtower as well.
> 
> What I'm trying to say is; can you elaborate a bit on why this is needed in watchtower as my immediate reaction that it is a bit out of scope?
> 
> Thanks 🙏🏼

hey, thanks for your thoughts, which are totally understandable for me.
Indeed, this behavior can be controlled by the docker flag `--memory`.  And this works fine for an end-user, which wants to control the memory usage of his container.
But as a platform provider having multiple consumers sharing a physical memory, it is important on runtime to check and control the memory consumption of the containers and if needed to limit that.

The key point is, developers may omit or misuse the docker flag, that can cause the whole platform to not perform correctly (Out Of Memory).

Since we are using watchtower to provide auto update of image, we thought it would make sense to use it for controlling the memory consumption with the filter ability. 

I think, other PaaS Providers do face the same issue.

Cheers,

simonaronsson:
> > I've since had some time to go through your changes, and while I do appreciate your effort **a lot** I'm having a bit of a hard time understanding why the changes are needed.
> > Isn't this behavior already controlled by the docker flag `--memory`? In that case, I think it would be preferable to avoid duplicating the effort by adding similar control mechanics to watchtower as well.
> > What I'm trying to say is; can you elaborate a bit on why this is needed in watchtower as my immediate reaction that it is a bit out of scope?
> > Thanks 🙏🏼
> 
> hey, thanks for your thoughts, which are totally understandable for me.
> Indeed, this behavior can be controlled by the docker flag `--memory`. And this works fine for an end-user, which wants to control the memory usage of his container.
> But as a platform provider having multiple consumers sharing a physical memory, it is important on runtime to check and control the memory consumption of the containers and if needed to limit that.
> 
> The key point is, developers may omit or misuse the docker flag, that can cause the whole platform to not perform correctly (Out Of Memory).
> 
> Since we are using watchtower to provide auto update of image, we thought it would make sense to use it for controlling the memory consumption with the filter ability.
> 
> I think, other PaaS Providers do face the same issue.
> 
> Cheers,

Got it, thanks. Unfortunately, I don't really think it fits within the scope of the project. I'd recommend that you set up a separate small go binary to do those checks for you, or maintain that feature in a fork. To me, it sounds very much like a restriction you should enforce during creation rather than by restarting containers during runtime.

The use case of watchtower is quite clear, and I'd prefer not to alter that at this point but focus on providing the best experience possible for that.

Thank you for understanding, and hope you find a reasonable way forward. 🙏

Issue #160, opened 2018-01-27
Use local time for --schedule?


steve28:
If I try to have watchtower do its thing at 4am every day by using the:
`--schedule "0 0 4 * * *"`
It seems to be interpreting this as 0400 UTC.  Per the log, I get:
`time="2018-01-27T22:26:55Z" level=info msg="First run: 2018-01-28 04:00:00 +0000 UTC" `

Is there any way to have watchtower use local time for the schedule?

solipsist01:
`docker run --name=watchtower -d -v /etc/localtime:/etc/localtime -v /etc/timezone:/etc/timezone -v /var/run/docker.sock:/var/run/docker.sock v2tec/watchtower --cleanup --schedule '0 0 4 * * *' `

stffabi:
Setting TZ and mounting `/usr/share/zoneinfo` as mentioned in #145, should work as expected.

If you have any further questions about this, feel free to reopen the issue.

stffabi:
Since version `0.3.0` you could just set the `TZ` environment variable and you don't need to mount anything else.

PR #461, opened 2020-03-12
document the TZ environment variable

Germs2004:
This adds a description of the "TZ" environment variable to the documentation so users will know it exists and works with Watchtower.  I only learned of this variable by reading the comments on issue #160.


simonaronsson:

simonaronsson:
@all-contributors add @Germs2004 for docs

PR #471, opened 2020-03-25
Time Zone improvements

auanasgheps:
Clarify where to find values for Time Zones.
I found myself into trouble and I thought this could be helpful!


simonaronsson:
Thank you very much! 🙏🏻

PR #483, opened 2020-03-27
Copy note about setting both interval and schedule

sixcorners:


simonaronsson:

simonaronsson:
@all-contributors add @sixcorners for docs

PR #501, opened 2020-04-11
Update usage-overview.md

anuar45:
Tried to use option "--interval 30" and it was failing with error "unknown flag". But "interval=30" works.


piksel:
@anuar45 In case someone else has this problem, would you mind sharing the conclusion?

anuar45:
> @anuar45 In case someone else has this problem, would you mind sharing the conclusion?

Hi, let me doublecheck and I will make PR if true.

Issue #440, opened 2020-01-15
Multiple private registries


khor0:
I can't find this in the documentation, so I'll assume it isn't a feature.
---

It would be awesome to declare in the configuration file which containers are to be monitored and their respective registries and credentials.

The example in [the documentation](https://containrrr.github.io/watchtower/private-registries/#create_the_configuration_file_manually) is as follows:
```json
{
  "auths": {
    "<REGISTRY_NAME>": {
      "auth": "XXXXXXX"
      }
   }
}
```

My suggestion is to have an array of containers and an option to watch over the configured ones or all, defaulting to watch over all containers and the registry to DockerHub:

```json
{
  "watchAll": true,
  "containers": [
    {
      "name": "container-name",
      "interval": 10,
      "registry": {
        "name": "registry.gitlab.com",
        "username": "username",
        "password": "password"
      }
    }
  ]
}
```

simonaronsson:
Thank you for your report! 🙏🏼

Unfortunately, the configuration file mentioned in the documentation is part of docker rather than watchtower, so such a change would require changes in moby/moby as well.

Is there a specific use case you're trying to solve that is currently not covered by watchtower? If so, it would be interesting to check whether we could find an alternative approach to achieve what you need.

khor0:
Hello there @simskij. Thanks for the fast response.

I have a server which contains multiple dockerized applications owned my multiple people/project groups.

As a sysadmin, I would rather have a Key on the server end which has read rights for the registry (in this case GitLab's) than have SSH keys deployed into the repo to execute commands inside the server.

I already use a simple instance of Watchtower with a self-hosted GitLab deployment. As it works very well, I would like to differentiate multiple registries for the projects and also 'attachments', such as databases.

If I my explanation was confusing, feel free to say so!

Thanks again

simonaronsson:
Got it! Thank you for the explanation!

For the time being I, unfortunately, have no plans on adding such a feature. This is more due to lack of time than anything else, so if you'd like to have a stab at providing a PR I'd be more than happy to assist you through the process.

The easiest way to do this would likely be to add a `WATCHTOWER_SCOPE_UID` env var, to be able to differentiate between different watchtower instances, and then set multiple instances with different scopes and keys. It would also have the added benefit of being a lot more resilient than just adding programmatic support for one instance to use multiple configs.

PR #511, opened 2020-04-23
Adds scopeUID config to enable multiple instances of Watchtower

### LINKED ISSUE #468: Multiple running watchtower instances
**Describe the bug**

I am implementing watchtower and I would like to know how I can do my 
case of use.

I have 2 types of containers: `[image]:test` and `[image]:prod`.

I would like the containers with tag `test` to be updated every 5 minutes. And the containers with tag `prod` through a schedule at 5am.

After studying it, I have decided to implement 2 watchtower:
 - The first with name `watchtower-test` with args:
	```--cleanup --tlsverify --interval 300 --label-enable```
 - The second with name `watchtower-prod` with args:
	```--cleanup --tlsverify --schedule "0 0 5 * * *"```

And add to `[image]:test` containers the label:
```com.centurylinklabs.watchtower.enable = true```

But I just realized that there cannot be 2 active watchtower containers.
[see code](https://github.com/containrrr/watchtower/blob/3fb299794b54175e5f06a6e12694651b596d5a4e/internal/actions/check.go#L36)

Could this deletion be parameterized?
Is there another possible solution? What do you think?

**To Reproduce**
```bash
$ docker run -d --name watchtower-prod --label com.centurylinklabs.watchtower.enable=false -v /var/run/docker.sock:/var/run/docker.sock containrrr/watchtower:0.3.10
[some_id1]

$ docker run -d --name watchtower-test --label com.centurylinklabs.watchtower.enable=false -v /var/run/docker.sock:/var/run/docker.sock containrrr/watchtower:0.3.10
[some_id2]

$ docker ps --all --quiet --filter name='watchtower$'|wc -l
1
```

Regards,

victorcmoura:
This pull request contains a Watchtower scope UID config, mentioned in #468 and #440.

With scopeUIDs, there can be multiple Watchtower running instances and each of them will monitor the containers within its scope, identified by a label:

- com.centurylinklabs.watchtower.scope-uid

This feature should not affect users who don't explicitly set a scope. Whenever it is not set, cleaning up will happen and multiple instances won't be allowed.

To do:

- [x] Core implementation of the new argument (WATCHTOWER_SCOPE_UID)
- [x] Filter change to locate containers within a scope
- [x] Check change to skip container clean ups when a scope uid is set
- [x] Argument documentation
- [x] Usage documentation
- [x] Use case tests

closes #468
If you have any suggestions, please let me know =)


victorcmoura:
Given that this feature affects [filters' BuildFilter func](https://github.com/containrrr/watchtower/blob/3e069a707e56e97c951649c730e6e6189494a94f/pkg/filters/filters.go#L55) and some documentation had to be done, I added a further explanation on how Watchtower's filters work: https://github.com/containrrr/watchtower/pull/511/commits/99438f0fab123f415f641fd2456021c7e8e37e3c#diff-5552622f71ec3fbab480701afca15a18R29. That should prevent more issues like https://github.com/containrrr/watchtower/issues/519 from happening

docs/running-multiple-instances.md
```diff
@@ -0,0 +1,32 @@
...
+Notice that:
+- Multiple instances can run with the same scope;
+- An instance without a scope will clean up other running instances, even if they have a defined scope;
+
+To define an instance monitoring scope, use the `--scope-uid` argument or the `WATCHTOWER_SCOPE_UID` environment variable on startup and set the _com.centurylinklabs.watchtower.scope-uid_ label with the same value for the containers you want to include in this instance's scope.
+
+For example, in a Docker Compose config file:
+
+```json
+version: '3'
```
zoispag: Is this a new file @victorcmoura ?
If yes, you also need to add it to the docs menu, by editing `mkdocs.yml`

victorcmoura: Yes it is! I'll add it there

internal/actions/check.go
```diff
@@ -33,7 +34,12 @@ func CheckForMultipleWatchtowerInstances(client container.Client, cleanup bool)
 		return nil
 	}
 
-	log.Info("Found multiple running watchtower instances. Cleaning up.")
+	if scopeUID != "" {
```
simonaronsson: If there are multiple watchtower containers running using the same scope UID, we should probably clean them up.

internal/flags/flags.go
```diff
@@ -125,6 +125,12 @@ func RegisterSystemFlags(rootCmd *cobra.Command) {
 		"",
 		viper.GetString("WATCHTOWER_HTTP_API_TOKEN"),
 		"Sets an authentication token to HTTP API requests.")
+	
+	flags.StringP(
```
simonaronsson, line 138: I think we could drop the `-uid` part of this flag, as well as for the env var.

pkg/container/metadata.go
```diff
@@ -6,6 +6,7 @@ const (
 	enableLabel           = "com.centurylinklabs.watchtower.enable"
 	dependsOnLabel        = "com.centurylinklabs.watchtower.depends-on"
 	zodiacLabel           = "com.centurylinklabs.zodiac.original-image"
+	scopeUID			  = "com.centurylinklabs.watchtower.scope-uid"
```
simonaronsson: Same here, `-uid` seems superfluous.

simonaronsson (changes requested):
Some minor things. Other than that, great work as usual!

internal/actions/check.go
victorcmoura: Instead of actually cleaning a running instance, would it be ok if it just didn't start? The implementation looks simpler this way.

simonaronsson: That would disable the auto-update feature of watchtower. I think the way to approach this is that, if scope is set, send a filter which also contains the scope on the line below, and then just proceed as usual

```
  containers, err := client.ListContainers(filters.WatchtowerContainersFilter)
```

victorcmoura: Changed the filter from
```
client.ListContainers(filters.WatchtowerContainersFilter)
```
to
```
client.ListContainers(filters.FilterByScope(scope, filters.WatchtowerContainersFilter))
```
But, since actual filtering is needed to test the logic of this function (and the mocked [`ListContainers`](https://github.com/containrrr/watchtower/blob/f76c48a95ec0bb61d84080a3546d529d8ea3f839/internal/actions/mocks/client.go#L43) simply returns all of the running containers. PS: changing this behavior causes multiple fatal failures in tests), I removed the test I had written in `actions_suite_test.go`.

But, as `FilterByScope` is being tested in `filters_test.go`, everything should be fine. But, in the future, we might need to put some effort in making a working mock of `ListContainers`.

docs/running-multiple-instances.md
```diff
@@ -0,0 +1,31 @@
+By default, Watchtower will clean up other instances and won't allow multiple instances monitoring the same containers. It is possible to override this behavior by defining a [scope](https://containrrr.github.io/watchtower/arguments/#filter_by_scope) to each running instance. 
+
+Notice that:
+- Multiple instances can run with the same scope;
```
simonaronsson: This should be `can't` right? :)

docs/container-selection.md
```diff
@@ -23,3 +23,9 @@ Or, it can be specified as part of the `docker run` command line:
 ```bash
 docker run -d --label=com.centurylinklabs.watchtower.enable=true someimage
 ```
+
+If you wish to create a monitoring scope, you will need to [run multiple instances and set a scope for each of them](https://containrrr.github.io/watchtower/running-multiple-instances).
+
+Watchtower filters running containers by testing them against each configured criteria. A container is monitored if all criteria are met. For example:
+- If a container's name it's on the monitoring name list (not empty `--name` argument) but it is not enabled (_centurylinklabs.watchtower.enable=false_), it won't be monitored;
+- If a container's name it's not on the monitoring name list (not empty `--name` argument), even if it is enabled (_centurylinklabs.watchtower.enable=true_ and `--label-enable` flag is set), it won't be monitored;
```
simonaronsson: `If a container's name it's` -> `If a container's name is`

docs/running-multiple-instances.md
```diff
@@ -0,0 +1,31 @@
+By default, Watchtower will clean up other instances and won't allow multiple instances monitoring the same containers. It is possible to override this behavior by defining a [scope](https://containrrr.github.io/watchtower/arguments/#filter_by_scope) to each running instance.
```
simonaronsson: `won't allow multiple instances monitoring the same containers` -> `won't allow multiple instances running on the same docker host or swarm` maybe?

docs/running-multiple-instances.md
```diff
@@ -0,0 +1,31 @@
...
+      - "com.centurylinklabs.watchtower.enable=true"
+      - "com.centurylinklabs.watchtower.scope=myscope"
+
+  watchtower:
+    image: containrrr/watchtower
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+    command: --debug --interval 30
+    labels:
+      - "com.centurylinklabs.watchtower.enable=false"
```
simonaronsson: What does this line do in this context? Disable self-update?

simonaronsson (changes requested):
All in all, this looks really good. 👍  Would you mind having a look at my remarks before I merge this?

docs/running-multiple-instances.md
victorcmoura: Yes it should. Just fixed it =)

docs/container-selection.md
victorcmoura: Oops, just missed it. Fixed.

docs/running-multiple-instances.md
victorcmoura: Sounds more correct. Fixed.

docs/running-multiple-instances.md
victorcmoura: Actually this whole example was pretty wrong:
- Port binding even though API mode is not enabled
- Use of `com.centurylinklabs.watchtower.enable` without passing `--label-enable`
- `--scope` is missing

I'm glad that you pointed it out so I could rewrite it =P

simonaronsson:
🥳🙏🏼

simonaronsson (approved):

PR #515, opened 2020-04-25
feat: add template support for shoutrrr notifications

### LINKED ISSUE #502: Don't add timestamps to shoutrrr notifications
One feature I’m missing while testing the new notifications on latest-dev is to disable the timestamps. Since I think most of the services show timestamps anyway, I think there should be an option to disable and/or format them in watchtower.

_Originally posted by @mbrandau in https://github.com/containrrr/watchtower/issues/264#issuecomment-612574815_

arnested:
In https://github.com/containrrr/watchtower/issues/264#issuecomment-612574815 @mbrandau suggested to make timestamp configurable in shoutrrr notifications. The idea has been added as a feature request in https://github.com/containrrr/watchtower/issues/502.

In https://github.com/containrrr/watchtower/issues/264#issuecomment-616231162 I suggested to maybe introduce a template for the messages.

This change introduces a template as a configuration for shoutrrr  notifications.

~~It still missed documentation etc. But I wanted to see what you think of the idea before going ahead.~~

*edit by @simskij:*
resolves #502

### DIFF
--- a/docs/notifications.md
+++ b/docs/notifications.md
@@ -179,6 +179,14 @@ To send notifications via shoutrrr, the following command-line options, or their
 Go to [containrrr.github.io/shoutrrr/services/overview](https://containrrr.github.io/shoutrrr/services/overview) to learn more about the different service URLs you can use.
 You can define multiple services by space separating the URLs. (See example below)
 
+You can customize the message posted by setting a template.
+
+- `--notification-template` (env. `WATCHTOWER_NOTIFICATION_TEMPLATE`): The template used for the message.
+
+The template is a Go [template](https://golang.org/pkg/text/template/) and the you format a list of [log entries](https://pkg.go.dev/github.com/sirupsen/logrus?tab=doc#Entry).
+
+The default value if not set is `{{range .}}{{.Message}}{{println}}{{end}}`. The example below uses a template that also outputs timestamp and log level.
+
 Example:
 
 ```bash
@@ -187,5 +195,6 @@ docker run -d \
   -v /var/run/docker.sock:/var/run/docker.sock \
   -e WATCHTOWER_NOTIFICATIONS=shoutrrr \
   -e WATCHTOWER_NOTIFICATION_URL="discord://token@channel slack://watchtower@token-a/token-b/token-c" \
+  -e WATCHTOWER_NOTIFICATION_TEMPLATE="{{range .}}{{.Time.Format \"2006-01-02 15:04:05\"}} ({{.Level}}): {{.Message}}{{println}}{{end}}" \
   containrrr/watchtower
 ```
--- a/internal/flags/flags.go
+++ b/internal/flags/flags.go
@@ -259,6 +259,12 @@ Should only be used for testing.
 		viper.GetString("WATCHTOWER_NOTIFICATION_GOTIFY_TOKEN"),
 		"The Gotify Application required to query the Gotify API")
 
+	flags.StringP(
+		"notification-template",
+		"",
+		viper.GetString("WATCHTOWER_NOTIFICATION_TEMPLATE"),
+		"The shoutrrr text/template for the messages")
+
 	flags.StringArrayP(
 		"notification-url",
 		"",
--- a/pkg/notifications/shoutrrr.go
+++ b/pkg/notifications/shoutrrr.go
@@ -1,7 +1,10 @@
 package notifications
 
 import (
+	"bytes"
 	"fmt"
+	"text/template"
+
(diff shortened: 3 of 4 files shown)


mbrandau:
I like that! However I would suggest to default to only showing the message itself - without timestamp and log level. We could then add the current template to documentation for those who like it more verbose.

arnested:
A simpler default is probably preferable. For a start I just wanted to make it backwards compatible (although the shoutrrr version of watchtower has not been released yet so there really is no need).

I'll go on with this later today.

arnested:
Oh. And thank you for the feedback, @mbrandau.

arnested:
Default template is changed to just message as suggested by @mbrandau and some documentation is added.

mbrandau (approved):
Looks good! If @simskij approves, #502 can be closed if this is merged.

simonaronsson:
Agreed. 👍

simonaronsson:
Would you mind adding two test cases; one for when the template is set and one for when it's not, and thus uses the default? Other than that, looks good to me 👍

simonaronsson:
Oh, another thing that came to mind: if the user supplies an invalid template, what would happen? Should we print a warning and fallback to the default? I personally think that would make sense to avoid disruptions to the service.

arnested:
Yes, makes sense. I'll do that.

simonaronsson:
Awesome! When this is all merged and done I'll go ahead and release `1.0.0` ✨ 🎉 🥂

arnested:
Awesome too, @simskij. It might be a couple of days until I have time to look into this, though.

simonaronsson:
No rush! 👍

arnested:
I just pushed a new version of this.

If the configured template is invalid it will fallback to the default template (and log an error).

Also added three test cases testing: 1) The default template when no logging configured, 2) a custom configured template, and 3) that the default template is used when configuring an invalid template.

go.sum
```diff
@@ -63,7 +63,6 @@ github.com/docker/distribution v2.7.1+incompatible h1:a5mlkVzth6W5A4fOsS3D2EO5BU
 github.com/docker/distribution v2.7.1+incompatible/go.mod h1:J2gT2udsDAN96Uj4KfcMRqY0/ypR+oyYUYmja8H+y+w=
 github.com/docker/docker v0.0.0-20190404075923-dbe4a30928d4 h1:34LfsqlE2kEvmGP9qbRoPvOWkmluYGzmlvWVTzwvT0A=
 github.com/docker/docker v0.0.0-20190404075923-dbe4a30928d4/go.mod h1:eEKB0N0r5NX/I1kEveEz05bcu8tLC/8azJZsviup8Sk=
-github.com/docker/docker v1.13.1 h1:IkZjBSIc8hBjLpqeAbeE5mca5mNgeatLHBy3GO78BWo=
```
simonaronsson: Is this deletion really supposed to be here?

go.sum
```diff
@@ -297,7 +296,6 @@ golang.org/x/net v0.0.0-20190311183353-d8887717615a/go.mod h1:t9HGtf8HONx5eT2rtn
 golang.org/x/net v0.0.0-20190404232315-eb5bcb51f2a3/go.mod h1:t9HGtf8HONx5eT2rtn7q6eTqICYqUVnKs3thJo3Qplg=
 golang.org/x/net v0.0.0-20190522155817-f3200d17e092 h1:4QSRKanuywn15aTZvI/mIDEgPQpswuFndXpOj3rKEco=
 golang.org/x/net v0.0.0-20190522155817-f3200d17e092/go.mod h1:HSz+uSET+XFnRR8LxR5pz3Of3rY3CfYBVs4xY44aLks=
-golang.org/x/net v0.0.0-20191116160921-f9c825593386 h1:ktbWvQrW08Txdxno1PiDpSxPXG6ndGsfnJjRRtkM0LQ=
```
simonaronsson: Is this deletion really supposed to be here?

simonaronsson (changes requested):
Looks great! Thanks a bunch! Two questions that I would like to cover, though, before we merge this.

go.sum
arnested: No, not related to the change.  Just a side effect of me having a habit of running `go mod tidy`. I'll push a commit without.

go.sum
arnested: No, not related to the change.  Just a side effect of me having a habit of running `go mod tidy`. I'll push a commit without.

simonaronsson:
Looks good to me! Thanks! 🎉

simonaronsson (approved):

PR #528, opened 2020-05-02
Increases stopContainer timeout to 10min

bopoh24:


simonaronsson:
Great catch! Didn't even notice that.
That means the update timeout until now has been 1 second, as that's the sleep duration of the timeout checker.

Thank you for the contribution!

simonaronsson:
@all-contributors add @bopoh24 for code and bugs

PR #537, opened 2020-05-09
Removed all potential debug password prints, both plaintext and encoded

### LINKED ISSUE #517: Watchtower logs the password from auth credentials to stdout in debug
**Describe the bug**
When watchtower is started in debug mode, it will log `Loaded auth credentials {username password host} from config.json`. I understand that debug mode is for debugging, but at the same time I don't think anyone ever wants their passwords logged to stdout if there's even a slightest chance of preventing it, and even in debug I probably wouldn't expect to see a password logged.

**To Reproduce**
1. Run watchtower in debug mode
2. Setup a private registry connection
3. Look in the stdout log
4. See login/password being logged

**Expected behavior**
I'd expect at least the password to be redacted from the output, and if logging it is critical for some reason, then make it explicit with an environment variable or a flag. Secure defaults are secure.

tammert:
Fixes #517.

Changed three debug messages: two in plaintext, one base64 encoded. This should prevent passwords being printed in debug logging.

Debug logging before fix:
```
INFO[0000] Running a one time update.                   
DEBU[0000] Checking containers for updated images       
DEBU[0000] Retrieving running containers                
DEBU[0000] Pulling localhost:5000/alpine for /fervent_roentgen 
DEBU[0000] Loaded auth credentials {testuser testpassword} for localhost:5000/alpine 
DEBU[0000] Got image name: localhost:5000/alpine        
DEBU[0000] Got auth value: eyJ1c2VybmFtZSI6InRlc3R1c2VyIiwicGFzc3dvcmQiOiJ0ZXN0cGFzc3dvcmQifQ==
```
Debug logging after fix:
```
INFO[0000] Running a one time update.
DEBU[0000] Checking containers for updated images
DEBU[0002] Retrieving running containers
DEBU[0002] Pulling localhost:5000/alpine for /fervent_roentgen
DEBU[0004] Loaded auth credentials for user testuser on registry localhost:5000/alpine
DEBU[0004] Got image name: localhost:5000/alpine
DEBU[0004] Got auth value
```

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -13,10 +13,10 @@ func GetPullOptions(imageName string) (types.ImagePullOptions, error) {
 		return types.ImagePullOptions{}, err
 	}
 
-	log.Debugf("Got auth value: %s", auth)
 	if auth == "" {
 		return types.ImagePullOptions{}, nil
 	}
+	log.Debugf("Got auth value")
 
 	return types.ImagePullOptions{
 		RegistryAuth:  auth,
--- a/pkg/registry/trust.go
+++ b/pkg/registry/trust.go
@@ -36,7 +36,7 @@ func EncodedEnvAuth(ref string) (string, error) {
 			Username: username,
 			Password: password,
 		}
-		log.Debugf("Loaded auth credentials %s for %s", auth, ref)
+		log.Debugf("Loaded auth credentials for user %s on registry %s", auth.Username, ref)
 		return EncodeAuth(auth)
 	}
 	return "", errors.New("Registry auth environment variables (REPO_USER, REPO_PASS) not set")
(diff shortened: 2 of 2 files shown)


victorcmoura:
Hi, @tammert! Thank you for your contribution. It looks good and certainly sets a better default behavior. The only thing that I believe would be handy is a way to actually show the auth value, as mentioned in #517:

> if logging it is critical for some reason, then make it explicit with an environment variable or a flag.

Since it is a particular use case, let me know if you don't feel up to implementing it, we can create a new issue and let it for future contributions :)

simonaronsson:
Let's create a separate issue for that. This solves the most immediate issue. I'll go ahead and merge this.
If you feel up for it, @tammert, feel free to go ahead and do that too and submit it in a separate PR :)

Thanks again for the contribution!

simonaronsson:
@all-contributors add @tammert for code

Issue #535, opened 2020-05-08
Blacklist in addition to Whitelist


Miosame:
Currently there is a whitelist behavior you can enable via [filter_by_enable_label ](https://containrrr.github.io/watchtower/arguments/#filter_by_enable_label) but a lot of setups would benefit instead from setting a blacklist of containers by label, simply because there is so many containers that would need to get a label added, versus just a few that shouldn't be checked each time - e.g. custom locally built containers.

simonaronsson:
Hey,

Thank you for your issue!

That is already possible, by setting `com.centurylinklabs.watchtower.enable` to `false` without passing `--label-enable`.

Hope that helps :)

Miosame:
@simskij ~~so how would watchtower then decide what containers it _should_ update? the proposed issue I have is to just add a label to containers it should straight up ignore, basically the same behavior as enable - but disable.~~

So the containers that have `com.centurylinklabs.watchtower.enable` set to false will be ignored? if that's the case, that would be extremely helpful to mention in the docs.

victorcmoura:
@Miosame we are planning to mention it soon: https://github.com/containrrr/watchtower/pull/511#issuecomment-622283650

simonaronsson:
It is in the docs, although not that clearly described:

> By default, watchtower will watch all containers. However, sometimes only some containers should be updated.
>
> If you need to exclude some containers, set the com.centurylinklabs.watchtower.enable label to false.

https://containrrr.github.io/watchtower/container-selection/

Miosame:
I see, it was a separate page entirely, seeing that everything else was listed in Arguments too incl. examples and that was always the page I hit by googling things, I never went outside that page. Maybe a small note in the enable argument section linking back would help others alike to find it? :smile:

simonaronsson:
That would be great! Would you mind creating a PR? :) I'd be happy to help you getting it merged.

Miosame:
@simskij absolutely, I can see multiple ways of how this could be handled, one of these in #540

simonaronsson:
Closing this now that the PR is merged. Thank you for your contribution! 🙏

PR #540, opened 2020-05-09
Add blacklist behavior description

Miosame:
As discussed in #535, one of the ways I could see adding a note. Though if possible it would probably be nice to link back to `container-selection` as it will contain more information soon from #511.


simonaronsson:
Thank you for the contribution! 🎉

simonaronsson:
@all-contributors add @miosame for docs

PR #544, opened 2020-05-11
Adds the option to skip TLS verification for a Gotify instance

### LINKED ISSUE #382: Add flag to disable TLS check for all notification provider, or better allow certificate providing
The current release of Watchtower only features the disable TLS flag for the E-Mail notification provider, yet any other notification provider, where the host can be on-premise, may be locked behind a self-signed cert (Gotify in this case) and therefore would need the same treatment.

I would like the addition of the disable TLS flag for all notification providers globally, or at least for those that can be hosted on-premise.
Additionally, the ability to provide the self-signed certificate for the notification provider would be extra awesome! (so we don't have to disable TLS verification 😃 )

The current workaround is to bind-mount an altered `ca-certificates.crt` that includes the self-signed certificate.
Alternatively, if the notification service runs in a docker container behind a reverse proxy, then the service could be reached via the docker network without TLS. (untested but should work theoretically)

### Additional context
My motivation for this request is only my home-lab, yet I know that quite a few companies use self-signed certificates internally (including the one I work at), which could benefit from this functionality.

---

some very basic research that I have done on how this might be done in GO (not a GO dev ☹️):
* [Set Insecure](https://stackoverflow.com/a/12122718)
* [Make Client Trust Server](https://ericchiang.github.io/post/go-tls/#getting-the-client-to-trust-the-server)
* [Golang https examples](https://github.com/jcbsmpsn/golang-https-example)

*not sure why I include this, yet maybe this comes in handy, I don't know* 😕 

---

Anyway, hope I am not asking for too much 😟.
Besides that, thank you all for this great project and keep up the great work!

tammert:
Fixes #382.

Tested this with the following Gotify server using a self-signed certificate:
```
docker run -d -p 443:443 -v /tmp/gotify:/app/data -e GOTIFY_SERVER_SSL_ENABLED=true -e GOTIFY_SERVER_SSL_CERTFILE=/app/data/localhost.crt -e GOTIFY_SERVER_SSL_CERTKEY=/app/data/localhost.key gotify/server
```
With the new flag to the instance as described above:
```
$ ./watchtower --trace --notification-gotify-url=https://localhost:443 --notification-gotify-tls-skip-verify --notification-gotify-token=AL8lZtthYtxzSgw --notifications=gotify -R
INFO[0000] Running a one time update.
DEBU[0000] Checking containers for updated images
```
Without the new flag:
```
$ ./watchtower --trace --notification-gotify-url=https://localhost:443 --notification-gotify-token=AL8lZtthYtxzSgw --notifications=gotify --notifications-level=info -R
INFO[0000] Running a one time update.
Failed to send Gotify notification:  Post "https://localhost:443/message?token=AL8lZtthYtxzSgw": x509: certificate signed by unknown authority
DEBU[0000] Checking containers for updated images
```

And, of course, the notifications were received when the new flag was active :).

### DIFF
--- a/docs/notifications.md
+++ b/docs/notifications.md
@@ -170,6 +170,8 @@ docker run -d \
   containrrr/watchtower
 ```
 
+If you want to disable TLS verification for the Gotify instance, you can use either `-e WATCHTOWER_NOTIFICATION_GOTIFY_TLS_SKIP_VERIFY=true` or `--notification-gotify-tls-skip-verify`. 
+
 ### [containrrr/shoutrrr](https://github.com/containrrr/shoutrrr)
 
 To send notifications via shoutrrr, the following command-line options, or their corresponding environment variables, can be set:
--- a/internal/flags/flags.go
+++ b/internal/flags/flags.go
@@ -183,10 +183,8 @@ func RegisterNotificationFlags(rootCmd *cobra.Command) {
 		"notification-email-server-tls-skip-verify",
 		"",
 		viper.GetBool("WATCHTOWER_NOTIFICATION_EMAIL_SERVER_TLS_SKIP_VERIFY"),
-		`
-Controls whether watchtower verifies the SMTP server's certificate chain and host name.
-Should only be used for testing.
-`)
+		`Controls whether watchtower verifies the SMTP server's certificate chain and host name.
+Should only be used for testing.`)
 
 	flags.StringP(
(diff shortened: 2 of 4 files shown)


simonaronsson (approved):

simonaronsson:
Thank you for the contribution! The only check that's failing is godacov, so I'll go ahead and merge this regardless!

Issue #532, opened 2020-05-04
Does watchtower support setting password from docker secrets file?


kevdogg:
Due to some of my other containers now requiring sensitive data be  passed through docker secret variables, I was wondering if watchtower supported use of docker secrets -- specifically:

Is there an equivalent of
WATCHTOWER_NOTIFICATION_EMAIL_SERVER_PASSWORD=<password>  

with something similar to
WATCHTOWER_NOTIFICATION_EMAIL_SERVER_PASSWORD_FILE=/run/secret/watchtower_notification_email_server_password_secret

From my testing, it doesn't seem like my notation with the docker secrets file specified works as I get an error message from the container that username/password not accepted by the smtp server.

simonaronsson:
Hi @kevdogg,

Not currently, but I would be more than happy to assist with reviews and suggestions in case anyone would like to have a stab at providing a PR.

Best,
Simme

tammert:
I will look into this! @kevdogg, would the `/run/secret/watchtower_notification_email_server_password_secret` file contain the plaintext password or do you expect a different format there?

kevdogg:
It would contain the plaintext format.

tammert:
PR is up :partying_face:

kevdogg:
PR is up? Sorry I don't understand the meaning however I did checkout your link.  I believe this addition is still awaiting approval? I tried the following:

 ```
secrets:
  watchtower_notification_email_server_secret:
    file: ./watchtower_notification_email_server.secret

services:
  watchtower:
    image: containrrr/watchtower:latest
    ...
    ...
    environment:
      - WATCHTOWER_NOTIFICATION_EMAIL_SERVER_PASSWORD=/run/secrets/watchtower_notification_email_server_secret

```

but it didn't seem to work

tammert:
Hey @kevdogg! PR = Pull Request, which means I've uploaded my code for @simskij to look at and merge into the other code, if he agrees with what I've done. After that we'll still need to create a new version of the Docker image which actually uses that code before you can use it as-is.

You can also build a new Docker image yourself from the source code, but that might be a bit too technical.

simonaronsson:
This has now been merged into master. @kevdogg; If you want to try it out, it should be available in the `latest-dev` image, otherwise it will be included in the next release. 👍

kevdogg:
Thanks for the inclusion

piksel:
[...]

PR #545, opened 2020-05-12
Allows flags containing sensitive stuff to be passed as files

tammert:
Potential fix for #532.

I've built in a check to see if the following sensitive options are passed to Watchtower as a file instead of a plaintext secret:
* notification-email-server-password
* notification-slack-hook-url
* notification-msteams-hook
* notification-gotify-token
* ~notification-url~

If this is the case, the runtime value is converted to the contents of the file instead. As the potential for an unexpected conflict is really low (aka a plaintext password that's actually the same as a fully-qualified existing file on the system), it seemed a better option than adding 5 extra flags.

@simskij if you hate this idea and/or have better suggestions, let me know!

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -81,6 +81,7 @@ func PreRun(cmd *cobra.Command, args []string) {
 		scheduleSpec = "@every " + strconv.Itoa(interval) + "s"
 	}
 
+	flags.GetSecretsFromFiles(cmd)
 	cleanup, noRestart, monitorOnly, timeout = flags.ReadFlags(cmd)
 
 	if timeout < 0 {
--- a/docs/notifications.md
+++ b/docs/notifications.md
@@ -33,7 +33,7 @@ To receive notifications by email, the following command-line options, or their 
 - `--notification-email-server-tls-skip-verify` (env. `WATCHTOWER_NOTIFICATION_EMAIL_SERVER_TLS_SKIP_VERIFY`): Do not verify the TLS certificate of the mail server. This should be used only for testing.
 - `--notification-email-server-port` (env. `WATCHTOWER_NOTIFICATION_EMAIL_SERVER_PORT`): The port used to connect to the SMTP server to send e-mails through. Defaults to `25`.
 - `--notification-email-server-user` (env. `WATCHTOWER_NOTIFICATION_EMAIL_SERVER_USER`): The username to authenticate with the SMTP server with.
-- `--notification-email-server-password` (env. `WATCHTOWER_NOTIFICATION_EMAIL_SERVER_PASSWORD`): The password to authenticate with the SMTP server with.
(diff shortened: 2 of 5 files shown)


simonaronsson (changes requested):
I like it! I only have two things I'd like to address before approving:

1. Would you mind fixing the linting issue?
2. Would be great if we could have a test or two verifying this behaves as expected (pref. both for a file and for a string)

I also noticed it's in need of a rebase to resolve some conflicts :)

tammert:
@simskij thanks for the feedback! I've added tests and fixed the linting issue. While creating the tests I found out that `--notification-url` is actually an array (would not have expected that from the name, maybe `--notications-urls` would be better?). Because of that, I couldn't really get an elegant solution in place for passing it as a file, so I've removed it from that functionality for now (and updated docs).

simonaronsson:
> maybe `--notications-urls` would be better?

Now that it's out in production, as of 1.0.0, it's hard to change it without breaking peoples setups, but we could definitely alias it to `--notification-urls` as well, if that makes sense. 

> Because of that, I couldn't really get an elegant solution in place for passing it as a file, so I've removed it from that functionality for now (and updated docs).

Seems reasonable! I'll review as soon as I have some time on my hands. Great work, as usual 👍

simonaronsson (approved):

piksel:
Regarding the URLs, the reason for it being singular is probably because you would pass it multiple times for multiple values, and as such `--notification-url foo --notification-url bar` makes sense. But what was the problem with getting it to work elegantly?
My suggestions would be either pass `--notification-urls myurls.txt` which would load the URLs from a newline-separated file, or to use multiple `--notification-url myurl1.txt`  arguments, one for each file.

simonaronsson:
> Regarding the URLs, the reason for it being singular is probably because you would pass it multiple times for multiple values, and as such `--notification-url foo --notification-url bar` makes sense. But what was the problem with getting it to work elegantly?
> 
> My suggestions would be either pass `--notification-urls myurls.txt` which would load the URLs from a newline-separated file, or to use multiple `--notification-url myurl1.txt`  arguments, one for each file.

I think the issue was that while that approach works well for flags, it won’t work with env vars as they will just overwrite each other.

piksel:
Oh, I just figured that was part of anaconda/viper, but I guess it's a single argument for command line as well then?

Update: No, it is indeed viper magic: [spf13/viper/issues/339](https://github.com/spf13/viper/issues/339)

tammert:
So my main issues when working with this PR were:
1) finding out at the end of the road that `--notification-url` was not a String but a StringArray, which made it incompatible with the func I had written
2) I couldn't find a way to set the value of a StringArray in the same way as I could for a String: https://github.com/containrrr/watchtower/blob/12d323354f222d8c4a0b5b852f15a475d943cd2b/internal/flags/flags.go#L400

I'm sure we can figure out a proper solution to those two issues though. I'd be happy to give that some more attention, but it would indeed by good to track it in a new issue/PR.

PR #562, opened 2020-05-30
feat(logging): use human readable log format in stdout/docker logs

piksel:
When investigating errors I find it really hard to parse the `logfmt` format that `logrus` outputs whenever it detects a non-TTY output. 
Since the logs are probably most often viewed through `docker logs` it might be a good idea to always use the "human readable" format for logging to stdout.

This PR will make watchtower use the following format in `docker logs`:
```
INFO[2020-05-30T15:10:33+02:00] Starting Watchtower and scheduling first run: 2020-05-30 15:15:33 +0200 CEST m=+301.094569801
```
instead of
```
time="2020-05-30T15:10:33+02:00" level=info msg="Starting Watchtower and scheduling first run: 2020-05-30 15:15:33 +0200 CEST m=+301.094569801"
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -61,6 +61,8 @@ func Execute() {
 func PreRun(cmd *cobra.Command, args []string) {
 	f := cmd.PersistentFlags()
 
+	log.SetFormatter(&log.TextFormatter{ForceColors: true, FullTimestamp: true})
+
 	if enabled, _ := f.GetBool("debug"); enabled {
 		log.SetLevel(log.DebugLevel)
 	}


zoispag:
I don’t think this should be the default.
For example, if this is merged, it will break my filters in Kibana. Maybe conditionally with a flag?

simskij:
I kind of agree with @zoispag on this one. Maybe `--pretty-logs`?

simonaronsson (changes requested):
I kind of agree with @zoispag on this one. Maybe --pretty-logs?

piksel:
The problem with opt in is that its kind of hard to tell people to turn on a flag, reproduce their bug and submit again.
But yeah, I figured there was a reason for not doing it before. I guess having the option is better than nothing, though.

piksel:
I'd argue that this is superseded by #598.

PR #563, opened 2020-06-01
Update shoutrrr to get latest and updated services

arnested:


piksel:
Sorry for pushing to master and making you update this PR :) But it was only minor documentation changes.

arnested:
I just thought it would be nice getting it as much up to date as possible while waiting :) No problem.

PR #564, opened 2020-06-01
Make sure all shoutrrr notifications are sent

CedricFinance:
When watchtower is launched with --run-once, some notifications might not be sent because the app exits before the goroutine has sent all the notifications.


regystro:
Anyone with write access could review this? Thank you.

piksel (approved):

Issue #495, opened 2020-04-06
Ignoring list of containers to update when given as arguments


andyleadbetter:
**Describe the bug**
Have a suite of containers, some are created locally using dockerfile's and build, some are pulled from the docker hub registry.  I want watchtower to update the hub images only, and so launched the watchtower container with a list if container names to update, however watchtower is failing with an error about not finding an image that is associated to one of the local containers.

**To Reproduce**

Create two running containers, one using a local docker file, one using a docker hub image.
Explicitly name the containers 'Hub' and 'Local'

Launch watchtower container, with the argument 'Hub' 

I expect watchtower to monitor the hub container and update only that one. 

**Screenshots**

**Environment**
Debian 
Docker 19.03

**Logs from running watchtower with the `--debug` option**
time="2020-04-06T17:10:12Z" level=debug msg="Sleeping for a second to ensure the docker api client has been properly initialized."
time="2020-04-06T17:10:13Z" level=debug msg="Retrieving running containers"
time="2020-04-06T17:10:13Z" level=fatal msg="Error: No such image: sha256:b598110d0fffc42862269a25d8767dd95764d0740f318fd6c0a097f8a22de5bf"

**Additional context**

tammert:
So, I think Watchtower is actually correctly ignoring other images in this.

There is an ongoing bug which caused a crash when local images were missing (#215), which I believe caused what you're seeing too. After the proposed fix in #571 I'm confident this problem would be gone too!

simonaronsson:
Fixed by #571

PR #571, opened 2020-06-07
Image of running container no longer needed locally

### LINKED ISSUE #215: Erron on synology : No such image
Since the last update, I get this message : level=fatal msg="Error: No such image: sha256:a6dc451b8bc1df6a70d1fbd5f11f6fe04613eb1ef31acd891f12fed709fbf01c"

tammert:
Fixes #215 and #495.

Issue arose when a running container did not have the associated image available locally anymore. This was an issue because the sha256 imageID was extracted from the `imageInfo`. However, the sha256 imageID is already present in `containerInfo`, which eliminates the need to even grab the `imageInfo` for running containers.

With this fix, the imageID is now grabbed from `containerInfo`, and getting `imageInfo` from running containers has been removed as it is not needed.

### DIFF
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -119,12 +119,7 @@ func (client dockerClient) GetContainer(containerID string) (Container, error) {
 		return Container{}, err
 	}
 
-	imageInfo, _, err := client.api.ImageInspectWithRaw(bg, containerInfo.Image)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


simonaronsson:
Great catch! 👏  🎉 
LGTM

simonaronsson (approved):

PR #577, opened 2020-06-10
feat(shoutrrr): allow setting URLs from file

piksel:
Extend #545 to work with `stringArray` arguments

### DIFF
--- a/internal/flags/flags.go
+++ b/internal/flags/flags.go
@@ -390,16 +390,14 @@ func GetSecretsFromFiles(rootCmd *cobra.Command) {
 
 // getSecretFromFile will check if the flag contains a reference to a file; if it does, replaces the value of the flag with the contents of the file.
 func getSecretFromFile(flags *pflag.FlagSet, secret string) {
-	value, err := flags.GetString(secret)
-	if err != nil {
-		log.Error(err)
-	}
+	flag := flags.Lookup(secret)
+	value := flag.Value.String()
 	if value != "" && isFile(value) {
 		file, err := ioutil.ReadFile(value)
 		if err != nil {
 			log.Fatal(err)
 		}
-		flag := flags.Lookup(secret)
+
 		if flag.Value.Type() == "stringArray" {
 			rows := bytes.Split(file, []byte{'\n'})
 
--- a/internal/flags/flags.go
+++ b/internal/flags/flags.go
@@ -390,26 +390,40 @@ func GetSecretsFromFiles(rootCmd *cobra.Command) {
 
 // getSecretFromFile will check if the flag contains a reference to a file; if it does, replaces the value of the flag with the contents of the file.
 func getSecretFromFile(flags *pflag.FlagSet, secret string) {
+
 	flag := flags.Lookup(secret)
-	value := flag.Value.String()
-	if value != "" && isFile(value) {
-		file, err := ioutil.ReadFile(value)
+	if flags.Lookup(secret).Value.Type() == "stringArray" {
+
+		values, err := flags.GetStringArray(secret)
 		if err != nil {
-			log.Fatal(err)
+			log.Error(err)
 		}
 
-		if flag.Value.Type() == "stringArray" {
-			rows := bytes.Split(file, []byte{'\n'})
-
-			for _, row := range rows {
-				err = flags.Set(secret, strings.TrimSpace(string(row)))
+		for _, value := range values {
+			if value != "" && isFile(value) {
+				file, err := ioutil.ReadFile(value)
 				if err != nil {
-					log.Error(err)
+					log.Fatal(err)
+				}
+
+				flag.Changed = false
+				rows := bytes.Split(file, []byte{'\n'})
+
+				for _, row := range rows {
+					err = flags.Set(secret, strings.TrimSpace(string(row)))
+					if err != nil {
+						log.Error(err)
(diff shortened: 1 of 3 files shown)


tammert:
Looks good on first glance, I'll try to give it a proper look and test soon! Only thing I'm not seeing here are updated docs, maybe you could include that as well? Especially if there's extra logic like 'one row per notification channel'.

piksel:
Yeah, you're right. Adding docs.

simonaronsson (approved):
LGTM

internal/flags/flags.go
```diff
@@ -388,19 +390,45 @@ func GetSecretsFromFiles(rootCmd *cobra.Command) {
...
-	if err != nil {
-		log.Error(err)
-	}
-	if value != "" && isFile(value) {
-		file, err := ioutil.ReadFile(value)
-		if err != nil {
-			log.Fatal(err)
-		}
-		err = flags.Set(secret, strings.TrimSpace(string(file)))
+
```
tammert, line 393: Maybe remove this whitespace?

internal/flags/flags.go
```diff
@@ -388,19 +390,45 @@ func GetSecretsFromFiles(rootCmd *cobra.Command) {
...
+			if value != "" && isFile(value) {
+				file, err := ioutil.ReadFile(value)
+				if err != nil {
+					log.Fatal(err)
+				}
+
+				flag.Changed = false
+				rows := bytes.Split(file, []byte{'\n'})
+
+				for _, row := range rows {
```
tammert, line 412: When running locally, I see 2 problems with this:
1) if the file ends in an empty line (something I always do, for example), it will add "" to the StringArray which causes Watchtower to crash with the error `FATA[0000] Failed to initialize Shoutrrr notifications: error initializing router services: unknown service scheme ''`
2) Even if I **don't** add an extra empty line at the end of the file, I'm still seeing that error. When debugging, I see the following (with two lines in the file):
image
Seems to be caused by how byte.Split() works.

Long story short: I think you'll need to add a `len(row) != 0` to make sure this works :)

internal/flags/flags.go
```diff
@@ -388,19 +390,45 @@ func GetSecretsFromFiles(rootCmd *cobra.Command) {
...
+					err = flags.Set(secret, strings.TrimSpace(string(row)))
```
tammert, line 413: Another issue: when using it as a file, you're only *adding* URLs, but you never remove the original value, aka the reference to the file. Causes the same error as above for me. Looks like this after the parsing is completed:
image

tammert (changes requested):

internal/flags/flags.go
piksel: There is something weird going on with the flags API. It should clear it, and they are cleared in the tests, but I need to investigate this further.

piksel:
There is a certain amount of jank with this method of setting the URLs. The "cleaner" solution would be to use another argument, like `--notification-url-file` or `--notification-urls` that is always a string and that appends the file contents.  
The problem is of course that it differs from how the other file loading arguments work (and having duplicates for all of them, with the `-file` suffix, seems really messy 🤔).
But perhaps it is less confusing compared to `--notification-url foo` => `load multiple urls from a file named foo`?

piksel:
The reason why this works from the tests but not from running is because of how viper is used in watchtower.
The way the authors meant it to be used is to override the viper configuration from the flags. What we instead do, is set the default values for the flags from the viper configuration (environment vars).
This means that overriding the array/slice values is not possible, as setting the value only appends it (unless it's set to it's default value, where it replaces it).

I have made two alterations to this, one is to move the default value to a `AddFlagValsFromEnv` method, in which values set from environment are updated using the `Flag.Set()` method and will append new entries when added from files. This still requires an additional argument (currently "--notification-url-file") as we cannot remove an entry from the flag values.

The other solution to this is to switch to using `viper.GetStringSlice("notification-url")` instead of `cmd.PersistentFlags().GetStringArray("notification-url")` to retrieve the configuration. This is how viper is meant to be used and the value can be replaced without using another argument. The only problem with doing this is that the rest of the arguments are still backwards.

simonaronsson:
@piksel 

I've dug around a bit in old changes, and it looks like you're right. The reason for this seems to be that when I moved from `urfave` to `cobra` and `viper`, I tried to do as few behavioral/logical changes as possible, likely forcing a pattern unfit for the new combination.

Old urfave code:
```go
  cli.StringFlag{	
    Name:   "host, H",	
    Usage:  "daemon socket to connect to",	
    Value:  "unix:///var/run/docker.sock",	
    EnvVar: "DOCKER_HOST",	
  },
```

Corresponding cobra+viper code:
```go
flags.StringP(
  "host", 
  "H", 
  viper.GetString("DOCKER_HOST"), 
  "daemon socket to connect to"
)
```

We should switch this around and use cobra as an input source bound to viper, and viper as the source of truth. That is more than slightly out of scope for this PR, but I still think it would be beneficial to do either together, or prior to merging this.

simonaronsson (changes requested):
Either we need to switch cobra and viper around in a separate PR or here, but until that, this isn't really mergeable

piksel:
I will push alt. 2 to this branch at least

piksel:
May be reopened (or replaced) when #684 is merged

PR #578, opened 2020-06-13
Removed accidental dot

tammert:
Noticed an extra `.` after `pkg/notifications/shoutrrr.go`, pretty sure it shouldn't be there (unless that is some magic syntax I don't know about 😅)


simonaronsson:
That's 100% accidental and a result of double spaces becoming `. ` on macOS 😅
Btw, if there are any parts of watchtower you feel you want to take ownership for, feel free to provide a PR adding yourself. 👍

PR #585, opened 2020-06-25
Add argument for setting priority level of Gotify notifications

bdr99:
I noticed that the priority setting for Gotify notifications was hardcoded as 0. This PR adds an argument, `--notification-gotify-priority` or `WATCHTOWER_NOTIFICATION_GOTIFY_PRIORITY`, to allow the user to set the priority. I kept 0 as the default value.

### DIFF
--- a/docs/notifications.md
+++ b/docs/notifications.md
@@ -172,6 +172,8 @@ docker run -d \
 
 `-e WATCHTOWER_NOTIFICATION_GOTIFY_TOKEN` or `--notification-gotify-token` can also reference a file, in which case the contents of the file are used.
 
+To specify the priority level for the Gotify notifications, you can use the `WATCHTOWER_NOTIFICATION_GOTIFY_PRIORITY` environment variable or the `--notification-gotify-priority` flag.
+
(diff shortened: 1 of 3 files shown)


victorcmoura (changes requested):
Thank you for your contribution! =)

The implementation itself looks good, due to the default value, it shouldn't affect running instances.

I only missed some instructions in the arguments documentation section. Would you mind adding your contribution to [arguments doc](https://github.com/containrrr/watchtower/blob/master/docs/arguments.md) as well?

bdr99:
@victorcmoura It doesn't look like any of the other notification-related arguments are mentioned in the arguments doc. For example, I don't see `--notification-gotify-url`, `--notification-email-to`, or even `--notifications` documented there. I think this omission is outside the scope of this PR.

victorcmoura:
Well pointed @bdr99, there really isn't any argument docs in this file. But, since it is actually an argument, there should be no reason not to document them.

@simskij can you help us with this? Is there any reason not to document this arguments?

simonaronsson:

Unfortunately, as all watchtower-native notification services will soon be replaced by shoutrrr adapters, I suggest you instead have a stab at implementing this there, as part of the URL: https://github.com/containrrr/shoutrrr

Merging this code would only lead to the feature being available for a short while and then disappearing again.

Issue #592, opened 2020-07-05
Common string functions in shoutrrr templates


PssbleTrngle:
**Is your feature request related to a problem? Please describe.**
The current shoutrrr template parsing does not allow basic string functions like lowercase, uppercase or capitalize

**Describe the solution you'd like**
I created a pull request with a possible solution, making the usage of these possible as string pipes

PR #593, opened 2020-07-05
Added string functions for lowercase, uppercase and capitalize to shoutrrr templates

PssbleTrngle:
Proposal for #592

### DIFF
--- a/pkg/notifications/shoutrrr.go
+++ b/pkg/notifications/shoutrrr.go
@@ -114,9 +114,7 @@ func getShoutrrrTemplate(c *cobra.Command) *template.Template {
 		"ToUpper": func(input interface{}) string {
 			return strings.ToUpper(fmt.Sprintf("%v", input))
 		},
-		"ToLower":func(input interface{}) string {
-			return strings.ToLower(fmt.Sprintf("%v", input))
-		},
+		"ToLower": strings.ToLower,
 		"Title":func(input interface{}) string {
 			return strings.Title(fmt.Sprintf("%v", input))
 		},
--- a/pkg/notifications/shoutrrr.go
+++ b/pkg/notifications/shoutrrr.go
@@ -111,9 +111,7 @@ func getShoutrrrTemplate(c *cobra.Command) *template.Template {
 	tplString, err := flags.GetString("notification-template")
 
 	funcs := template.FuncMap{
-		"ToUpper": func(input interface{}) string {
-			return strings.ToUpper(fmt.Sprintf("%v", input))
-		},
+		"ToUpper": strings.ToUpper,
 		"ToLower": strings.ToLower,
 		"Title":func(input interface{}) string {
 			return strings.Title(fmt.Sprintf("%v", input))
--- a/pkg/notifications/shoutrrr.go
+++ b/pkg/notifications/shoutrrr.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"text/template"
+    "strings"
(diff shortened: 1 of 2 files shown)


pkg/notifications/shoutrrr.go
```diff
@@ -109,10 +110,22 @@ func getShoutrrrTemplate(c *cobra.Command) *template.Template {
...
+	funcs := template.FuncMap{
+		"ToUpper": func(input interface{}) string {
+			return strings.ToUpper(fmt.Sprintf("%v", input))
+		},
+		"ToLower":func(input interface{}) string {
+			return strings.ToLower(fmt.Sprintf("%v", input))
+		},
+		"Title":func(input interface{}) string {
+			return strings.Title(fmt.Sprintf("%v", input))
+		},
```
piksel: It can be done much simpler.
```suggestion
		"Title": strings.Title,
```
This was actually the [example from the documentation](https://golang.org/pkg/text/template/#example_Template_func) 😁

pkg/notifications/shoutrrr.go
```diff
@@ -109,10 +110,22 @@ func getShoutrrrTemplate(c *cobra.Command) *template.Template {
 
 	tplString, err := flags.GetString("notification-template")
 
+	funcs := template.FuncMap{
+		"ToUpper": func(input interface{}) string {
+			return strings.ToUpper(fmt.Sprintf("%v", input))
+		},
```
piksel:
```suggestion
		"ToUpper": strings.ToUpper,
```

pkg/notifications/shoutrrr.go
```diff
@@ -109,10 +110,22 @@ func getShoutrrrTemplate(c *cobra.Command) *template.Template {
...
+		"ToLower":func(input interface{}) string {
+			return strings.ToLower(fmt.Sprintf("%v", input))
+		},
```
piksel:
```suggestion
		"ToLower": strings.ToLower,
```

PssbleTrngle:
That was my first try to, but if I give the `ToUpper` function an argument of another type than stirng, the log level for example, an error is thrown. That's why I first convert the input to string in each function

piksel:
Yeah, but you could just pipe it to `printf "%v"` in those cases. I'm not sure it should be up to the template functions to format non-strings.

simonaronsson:
I agree with @piksel

PssbleTrngle:
You are right, I commited your suggestions

pkg/notifications/shoutrrr_test.go
```diff
@@ -51,6 +51,30 @@ func TestShoutrrrTemplate(t *testing.T) {
 	require.Equal(t, "info: foo bar\n", s)
 }
 
+func TestShoutrrrStringFunctions(t *testing.T) {
+	cmd := new(cobra.Command)
+	flags.RegisterNotificationFlags(cmd)
+	err := cmd.ParseFlags([]string{"--notification-template={{range .}}{{.Level | ToUpper }}: {{.Message | ToLower }} {{.Message | Title }}{{println}}{{end}}"})
```
piksel:
```suggestion
	err := cmd.ParseFlags([]string{"--notification-template={{range .}}{{.Level | printf "%v" | ToUpper }}: {{.Message | ToLower }} {{.Message | Title }}{{println}}{{end}}"})
```
I think this ought to fix the failed test, but I haven't tried it.

piksel (approved):
LGTM!

PR #594, opened 2020-07-06
Warning if `WATCHTOWER_NO_PULL` and` WATCHTOWER_MONITOR_ONLY` are used simultaneously.

### LINKED ISSUE #570: Warn if using both `WATCHTOWER_NO_PULL` and `WATCHTOWER_MONITOR_ONLY` at the same time
> Maybe we should warn/fatal on doing this combination of options, since Watchtower will essentially be useless with both those specified?

_Originally posted by @piksel in https://github.com/containrrr/watchtower/discussions/557#discussioncomment-23224_

As there is a very very corner use case for this, where one builds images locally using some kind of automation, but dont want the update to be automatic, it should likely not error out but rather warn that this combination might be unintentional, preferably once at startup.

m-sedl:
A warning was added if `WATCHTOWER_NO_PULL` and` WATCHTOWER_MONITOR_ONLY` are used simultaneously. Closes #570

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -102,6 +102,11 @@ func PreRun(cmd *cobra.Command, args []string) {
 	reviveStopped, _ := f.GetBool("revive-stopped")
 	removeVolumes, _ := f.GetBool("remove-volumes")
 
+	if monitorOnly && noPull {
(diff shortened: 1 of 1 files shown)


simonaronsson:
Thanks for that! I went ahead and modified the message a bit, but other than that, this is more than welcome! 🙇

simonaronsson:
@all-contributors add @MihailITPlace for code

PR #598, opened 2020-07-08
Disabling color through environment variables

bugficks:
This patch adds support for disabling color through environment variables.

- NO_COLOR: https://no-color.org/
- CLICOLOR,CLICOLOR_FORCE: https://bixense.com/clicolors/ ([logrus built-in](https://github.com/sirupsen/logrus/blob/6699a89a232f3db797f2e280639854bbc4b89725/text_formatter.go#L46))

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -61,6 +61,18 @@ func Execute() {
 func PreRun(cmd *cobra.Command, args []string) {
 	f := cmd.PersistentFlags()
 
+	// https://no-color.org/
+	if _, enabled := os.LookupEnv("NO_COLOR"); enabled {
+		log.SetFormatter(&log.TextFormatter{
+			DisableColors: true,
+		})
+	} else {
+		// enable logrus built-in support for https://bixense.com/clicolors/
+		log.SetFormatter(&log.TextFormatter{
+			EnvironmentOverrideColors: true,
+		})
+	}
+
 	if enabled, _ := f.GetBool("debug"); enabled {
 		log.SetLevel(log.DebugLevel)
 	}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -61,8 +61,7 @@ func Execute() {
 func PreRun(cmd *cobra.Command, args []string) {
 	f := cmd.PersistentFlags()
 
-	// https://no-color.org/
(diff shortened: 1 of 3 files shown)


cmd/root.go
```diff
@@ -61,6 +61,18 @@ func Execute() {
...
+	// https://no-color.org/
+	if _, enabled := os.LookupEnv("NO_COLOR"); enabled {
+		log.SetFormatter(&log.TextFormatter{
+			DisableColors: true,
+		})
+	} else {
+		// enable logrus built-in support for https://bixense.com/clicolors/
+		log.SetFormatter(&log.TextFormatter{
+			EnvironmentOverrideColors: true,
+		})
```
piksel, line 73: Probably not a good idea, see #562

bugficks:
I have seen PR #562 but it changes rather important defaults while
this patch disables coloring only if user has set some environment
variable.
Without those env vars everything stays the same.

I could have added a command line param but I have opt-ed for some
standard way.
If this really isn't acceptable I might go that route though.

piksel:
Yeah, I noticed, so I removed the comment. I don't think there are any issues with this.

piksel (approved):

cmd/root.go
```diff
@@ -61,6 +61,18 @@ func Execute() {
 func PreRun(cmd *cobra.Command, args []string) {
 	f := cmd.PersistentFlags()
 
+	// https://no-color.org/
+	if _, enabled := os.LookupEnv("NO_COLOR"); enabled {
```
simonaronsson: Should be implemented the same way as all other flags and env vars.

simonaronsson (changes requested):
Thank you for your contribution! While I really like the feature you've added, I'd kindly like to ask you to implement it the same way as all other flags and env vars in watchtower, namely in https://github.com/containrrr/watchtower/blob/master/internal/flags/flags.go using viper and cobra. I'd also like to ask you to document this feature in https://github.com/containrrr/watchtower/blob/master/docs/arguments.md

Thanks again. 🙇

piksel (approved):

simonaronsson:
@all-contributors add @bugficks for code

simonaronsson:
Thank you for your contribution! 🙇

PR #610, opened 2020-07-31
lifecycle logs as Debug instead of Info

MichaelSp:
The life-cycle messages should not have the same level as the update messages.

#### Why?
I receive a (mail-) notification for every run even if no image is changed: 
```
2020-07-31 06:44:33 (info): Executing pre-check command.
```

But I only want to receive a notification if something changed:
```
2020-07-20 06:44:44 (info): Found new traefik:v2.2 image (sha256:7c992081f52423e325bd12debf0edcb4f7cfa398683c26f99c00112ec14e1ab9)
```

### DIFF
--- a/pkg/lifecycle/lifecycle.go
+++ b/pkg/lifecycle/lifecycle.go
@@ -36,7 +36,7 @@ func ExecutePreCheckCommand(client container.Client, container container.Contain
 		return
 	}
 
-	log.Info("Executing pre-check command.")
+	log.Debug("Executing pre-check command.")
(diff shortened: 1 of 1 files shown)


piksel (approved):
It's kind of hard to balance what should be printed in the default output and what should be included in the notifications, but I do agree with this one. I don't think it's necessarily helpful to print these messages when running without debug enabled.

simonaronsson:
Thanks for the contribution! 🙇

simonaronsson:
@all-contributors add @MichaelSp for code

PR #619, opened 2020-08-16
Rolling restart

osheroff:
bouncing each image individually can ensure that a group of docker containers launched with docker-compose can stay 100% up during deploy.   nice when used with a post-update hook to wait until the webserver is launched.


simonaronsson:
Great addition! Thank you for your contribution!

simonaronsson:
@all-contributors add @osheroff for code

Issue #649, opened 2020-09-19
Make watchtower skip update if pre-update lifecycle hook exits with a non-zero exit code


simonaronsson:
> Yeah, to be totally honest, I already thought we did this! 😅  This is the intended behavior, so we'll need to fix that. Thanks for reporting!

_Originally posted by @simskij in https://github.com/containrrr/watchtower/discussions/647#discussioncomment-81488_

simonaronsson:
This should be a small self-contained fix in `pkg/container/client.go` around line 380-383, where we need to add an error return:

```go
if execInspect.ExitCode > 0 {
    log.Errorf("Command exited with code %v.", execInspect.ExitCode)
    log.Error(execOutput)
}
```

ShaileshSurya:
Hi, @simskij Looks like a small change here. What should be the error to return here for exitcode other than 0? 

It will be great if you could provide the error string. 

Also, What is the timeline around this. This looks like targeted for hactoberfest and the event is still days away.

simonaronsson:
If you want to do it, feel free. I just thought it would be a good hacktoberfest issue, given the very limited scope. 😅 

The current error text, but lowercased and without a trailing dot will do for now. Switching that call to `fmt.Errorf` and then printing that one using the logger should be be sufficient. 👍

yrien30:
Is it possible to read message printed from the preupdate script and log that as error when exit code is higher than 0 ? I thought it would be saved to execOutput variable but that is not the case.

When WATCHTOWER_INCLUDE_STOPPED=true is set : then failing to run the script because the docker container is not running should not prevent a update.

simonaronsson:
> Is it possible to read message printed from the preupdate script and log that as error when exit code is higher than 0 ? I thought it would be saved to execOutput variable but that is not the case.
> 
> When WATCHTOWER_INCLUDE_STOPPED=true is set : then failing to run the script because the docker container is not running should not prevent a update.

Agreed, as long as it fails because it's stopped, and not because of some other arbitrary reason. We need to check this and make sure we don't ignore other reasons as well.

ryester19:
Can we do something like this for the exit codes?

0: Proceed with update
1: Skip update, don't log
2 and greater: Skip update, log as error

My reasoning for requesting this is because I want to use a script to prevent a pyload container from being updated and restarted while a download is still running. Having an error logged every 5 minutes would be an annoyance in my use case because these downloads can sometimes last 8 hours or more.

[...]

PR #651, opened 2020-09-25
Allow watchtower to update rebooting containers

### LINKED ISSUE #650: Updating containers that is rebooting
Hi,
if we at some point we introduce a problem in a docker container that cause crashing of the container then it will be standing in a reboot loop if --restart always is set on that container. Even if you include WATCHTOWER_INCLUDE_STOPPED=true then it will not update the container when it standing in a reboot loop. I though it would be a good idea to introduce a new setting WATCHTOWER_INCLUDE_REBOOTING so that when the docker container get fixed then it will pull down the new one. Restoring the operation of the docker container. 

rebooting status is not included in filter:
https://github.com/containrrr/watchtower/blob/98b518612b6ae2c005667de997bec67325cd98cf/pkg/container/client.go#L102-L112

yrien30:
Closes #650

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -31,7 +31,7 @@ var (
 	timeout        time.Duration
 	lifecycleHooks bool
 	rollingRestart bool
-	scope	         string
+	scope          string
 )
 
 var rootCmd = &cobra.Command{
@@ -116,6 +116,7 @@ func PreRun(cmd *cobra.Command, args []string) {
 
 	noPull, _ := f.GetBool("no-pull")
 	includeStopped, _ := f.GetBool("include-stopped")
+	includeRestarting, _ := f.GetBool("includeRestarting")
 	reviveStopped, _ := f.GetBool("revive-stopped")
 	removeVolumes, _ := f.GetBool("remove-volumes")
 
@@ -128,6 +129,7 @@ func PreRun(cmd *cobra.Command, args []string) {
 		includeStopped,
 		reviveStopped,
 		removeVolumes,
+		includeRestarting,
 	)
 
 	notifier = notifications.NewNotifier(cmd)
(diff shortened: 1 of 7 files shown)


yrien30:
Hi, was going to make a new pull request for #649. I wanted to have it in seperate pull request but since I did't start whit making a new branch before I started to make changes it came on same. I reverted the change for now. I will create new pull request for this when you have taken in the changes for #650.

piksel:
I just removed the referenced ticket in the title, because it messes with Githubs autocomplete for references

cmd/root.go
```diff
@@ -116,6 +116,7 @@ func PreRun(cmd *cobra.Command, args []string) {
 
 	noPull, _ := f.GetBool("no-pull")
 	includeStopped, _ := f.GetBool("include-stopped")
+	includeRestarting, _ := f.GetBool("includeRestarting")
```
piksel: Typo?
```suggestion
	includeRestarting, _ := f.GetBool("include-restarting")
```

piksel (changes requested):
This overall looks good to me.

The flag name doesn't align with the current ones though.

pkg/container/container_test.go
```diff
@@ -68,6 +69,18 @@ var _ = Describe("the container", func() {
...
+				client = dockerClient{
+					api:               docker,
+					pullImages:        false,
+					includeRestarting: true,
+				}
+				containers, err := client.ListContainers(filters.NoFilter)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(len(containers) > 0).To(BeTrue())
+			})
+		})
```
simonaronsson: could you please add a test case for the opposite as well? not including restarted when the flag is not set, that is.

simonaronsson (changes requested):

pkg/container/container_test.go
yrien30: Sure 👍

simonaronsson (approved):

yrien30:
Hi,  is there anything more to do on this or can it be merged now?

piksel (approved):

simonaronsson:
Thank you kindly for your contribution! 🙇

simonaronsson:
@all-contributors add @yrien30 for code

Issue #201, opened 2018-06-11
Watch without pulling the images?


radarsh:
I don't know if it's been previously asked or addressed. I would like to know if it's possible for watchtower to find out if an image has changed on the repository without having to download it, possibly using the registry API.

In its current state, watchtower keeps incrementing the pull counts on Docker Hub (which may not be such a big deal) but a more serious problem could be that it may exhaust the bandwidth on a limited bandwidth hosting set up if it downloads the entire image every few minutes.

lmcnulty4:
I'd like to know this as well. I don't really know Go but from poking around the source it looks like in the IsContainerStale method first downloads the image and then checks whether to update. Is it not possible to inspect the image without downloading and then download only if it's changed?

rbq:
I'd assume that only metadata (and updated layers) will be transferred? If the image is unchanged, the amount of data transferred should be minimal. Just like the second time you run `docker pull [image]` locally. I actually just tried that and caused 7.4 kB upstream and 29.8 kB downstream traffic. That is like … 320 MB per month and container or so, given a 5 minute interval?

That being said, I landed here because my Docker Hub counter went from 1 to 4000 over a weekend, which is a bit weird. Not sure if it is a Watchtower issue though—maybe instead Docker Hub should not count a pull w/o changes?

simonaronsson:
That's how docker hub works I'm afraid.
I've gone through the client api a bit as well and ImagePull currently seems to be our only option.

PR #674, opened 2020-11-08
add http head based digest comparison to avoid dockerhub rate limits

simonaronsson:
We currently experience a pretty urgent issue where each check by watchtower counts towards the DockerHub rate limit. What we've managed to understand so far after discussions with the docker OSS team and others:

- DockerHub does not count HEAD requests towards the rate limit
- Docker Engine was thought to do HEAD requests where possible. This, however, turned out to not be the case for various reasons outlined in https://github.com/moby/moby/pull/41607.
- Getting this fix out to the community will take time. My best bet is that even if it's released in 20.xx, it will take years until it reaches a significant portion of our users.
- The docker registry specification is fairly detailed in how registries should behave in terms of authentication, tokens, challenges, and what request headers do what.
- GHCR and DockerHub behave almost the same, which is a relief as they, together with quay, likely make up a significant part of all registry checks done by Watchtower.
- Depending on `skopeo` or any other repo in the `containers` org. is currently not an option. It has the same issue with fetching manifest list digests as docker engine, which would lead to digest mismatches for all multi-arch tags. As a multi-arch tag leads to at least 2 GET requests toward the rate limit, this in itself makes it a deal-breaker for us.
- I've not yet been able to verify whether HEAD requests bump the pull count. Hopefully, they don't, solving some other long-lingering issues (see #429, #201).
- To make sure we don't break anything for users on older/incompatible registries, we'll make sure to fallback to GET requests if a registry doesn't behave as expected.
- The Docker OSS team has kindly offered to work with us to see if there is anything to be done, in any end, to make the user experience smoother and less resource-consuming going forward. 

I'm fairly certain I've missed something, so we'll have to go through this again before releasing, but at least we're getting somewhere. 😅


pkg/registry/digest/digest_test.go
```diff
@@ -0,0 +1,64 @@
...
+	RepoDigests: []string {
+		"ghcr.io/k6io/operator@sha256:d68e1e532088964195ad3a0a71526bc2f11a78de0def85629beb75e2265f0547",
+	},
+}
+
+var (
+	DH_USERNAME = os.Getenv("CI_INTEGRATION_TEST_REGISTRY_DH_USERNAME")
+	DH_PASSWORD = os.Getenv("CI_INTEGRATION_TEST_REGISTRY_DH_PASSWORD")
+	GH_USERNAME = os.Getenv("CI_INTEGRATION_TEST_REGISTRY_DH_USERNAME")
+	GH_PASSWORD = os.Getenv("CI_INTEGRATION_TEST_REGISTRY_DH_PASSWORD")
```
piksel:
```suggestion
	GH_USERNAME = os.Getenv("CI_INTEGRATION_TEST_REGISTRY_GH_USERNAME")
	GH_PASSWORD = os.Getenv("CI_INTEGRATION_TEST_REGISTRY_GH_PASSWORD")
```

simonaronsson: Some refactoring inc. tomorrow. That change is in there, sort of! 😅

piksel: Yeah, no worries. My typo alarm just went off 😁
Felt like a typical problem one could spend way too long trying to figure out why the tests fail for no good reason.

dudipsh:
UP
@dudipsh

simonaronsson:
> UP
> @dudipsh

?

dudipsh:
> > UP
> > @dudipsh
> 
> ?

I just wanted to follow the solution :)

go.mod
```diff
@@ -53,7 +56,8 @@ require (
 	github.com/stretchr/testify v1.3.0
 	github.com/theupdateframework/notary v0.6.1 // indirect
 	github.com/zmap/zlint v1.0.2 // indirect
-	golang.org/x/net v0.0.0-20190522155817-f3200d17e092
+	golang.org/x/net v0.0.0-20200520004742-59133d7f0dd7
+	golang.org/x/sys v0.0.0-20201117170446-d9b008d0a637 // indirect
```
piksel:
```suggestion
	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9
	golang.org/x/sys v0.0.0-20190830141801-acfa387b8d69 // indirect
```
also works, and should lead to fewer side effects?
the only downside is ginko and gomega that cannot be upgraded past:
```
	github.com/onsi/ginkgo v1.11.0
	github.com/onsi/gomega v1.10.0
```
Gist of full go.mod: https://gist.github.com/piksel/03e87dcb7d4166ce5ab8a20ccf93aa9a

simonaronsson: What side effects are you thinking of?

piksel: The unknown ones :) The override makes it essentially just treat it as if it were another version, which it may not work correctly with.

itsmostafa (approved):

piksel (approved):

go.mod
simonaronsson: Fixed.

piksel: Looking at the specification, it seems like the override SHOULD work the same way as my solution. Why it insists on upgrading certain packages is beyond me.

pkg/registry/auth/auth.go
```diff
@@ -0,0 +1,168 @@
...
+
+	pairs := strings.Split(raw, ",")
+	values := make(map[string]string, 0)
+	for _, pair := range pairs {
+		trimmed := strings.Trim(pair, " ")
+		kv := strings.Split(trimmed, "=")
+		key := kv[0]
+		val := strings.Trim(kv[1], "\"")
+		values[key] = val
+	}
```
piksel: The values map could be initialized with `len(pairs)`

pkg/registry/auth/auth.go
```diff
@@ -0,0 +1,168 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/containrrr/watchtower/pkg/registry/helpers"
+	"github.com/containrrr/watchtower/pkg/types"
+	"github.com/docker/distribution/reference"
+	"github.com/sirupsen/logrus"
+	"io/ioutil"
+	"net/http"
+	url2 "net/url"
```
piksel: 🤨

mrmachine:
So, is this fixed now? in the `latest` image on Docker Hub? Does it fix merely by updating the default poll interval to 24h (from 5 minutes)? Or some other magic (head based digest comparison)?

scottrus:
> So, is this fixed now? in the `latest` image on Docker Hub? Does it fix merely by updating the default poll interval to 24h (from 5 minutes)? Or some other magic (head based digest comparison)?

Tag `latest-dev` on docker hub has this PR in it. This is not yet in `latest`. The PR is about using `HEAD` requests to avoid rate limits.

piksel:
Believe me, this:
image
is not just updating the poll interval 😁

simonaronsson:
@mrmachine Definitely some other magic. 😅

yrien30:
When will there be a new release of watchtower including this fix?

simonaronsson:
> When will there be a new release of watchtower including this fix?

As soon as we deem it stable enough. Thanks for your patience!

pneuschwander:
What was the reason behind the decision to change the default interval from 5 minutes to 24 hours? 
Is it related to the rate limit? 
Shouldn't this be solved by using HEAD requests?
Will I run into rate limit problems if I change back to 5 minutes?

Would be great if someone could give some background information about the decision / necessity to change the interval 🙂

simonaronsson:
> What was the reason behind the decision to change the default interval from 5 minutes to 24 hours?
> Is it related to the rate limit?
> Shouldn't this be solved by using HEAD requests?
> Will I run into rate limit problems if I change back to 5 minutes?
> 
> Would be great if someone could give some background information about the decision / necessity to change the interval 🙂

**TL;DR**

- It really didn't make sense to poll that aggressively given the release cadence of most public DockerHub images.
- It should be safe to set a tighter interval, but we still encourage you to
  - check your logs to make sure it works, and
  - ask yourself whether you really need to.

---

The decision was made for multiple reasons:

- We don't want to put unnecessary pressure on DockerHubs servers, regardless of whether it's microscopic in comparison to what they handle in total.
- Once every 24 hours should be sufficient for the users that only want a drop-in solution for keeping their home server up to date. Most external tools don't really change that frequently in any case.
- Users who **actually** need a tighter interval will still be able to set it themselves.

Whether or not you'll be hitting the limit really depends on how often the images you use actually change. While the head requests not count towards the rate limit if nothing is changed, pulling images still does.

So, if the image changes are frequent enough that the rate limit is hit because of actual pulls, then you would indeed still be able to hit the limit. If that's the case, you should probably pay DockerHub the 5 bucks a month they're asking for a pro account anyway. 😅

PR #684, opened 2020-11-22
feat(config): swap viper and cobra for config

### LINKED ISSUE #625: Switch around viper and cobra
Quote from #577:
> I've dug around a bit in old changes, and it looks like you're right. The reason for this seems to be that when I moved from `urfave` to `cobra` and `viper`, I tried to do as few behavioral/logical changes as possible, likely forcing a pattern unfit for the new combination.
>
> Old urfave code:
> ```go
>  cli.StringFlag{	
>    Name:   "host, H",	
>    Usage:  "daemon socket to connect to",	
>    Value:  "unix:///var/run/docker.sock",	
>    EnvVar: "DOCKER_HOST",	
>  },
>```
>
> Corresponding cobra+viper code:
> ```go
>flags.StringP(
>  "host", 
>  "H", 
>  viper.GetString("DOCKER_HOST"), 
>  "daemon socket to connect to"
>)
>```
>
>We should switch this around and use cobra as an input source bound to viper, and viper as the source of >truth. That is more than slightly out of scope for this PR, but I still think it would be beneficial to do either >together, or prior to merging this.

piksel:
This swaps the order of `spf13/viper` and `spf13/cobra` so that `viper` is the "main" configuration handler and `cobra`s flags is used to override the configuration.
Using them this way is the intended way by the authors and removes a lot of boiler plate code that was necessary before.

This also paves the way for more typed configuration and allows for fewer parameters to methods and less global vars.
More could be done on that front than what is included in this PR, as I choose to limit this to the intended scope.

Fixes #625

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,9 +1,10 @@
 package cmd
 
 import (
+	"fmt"
+	"github.com/spf13/viper"
 	"os"
 	"os/signal"
-	"strconv"
 	"syscall"
 	"time"
 
@@ -21,17 +22,9 @@ import (
 )
 
 var (
-	client         container.Client
-	scheduleSpec   string
-	cleanup        bool
-	noRestart      bool
-	monitorOnly    bool
-	enableLabel    bool
-	notifier       *notifications.Notifier
-	timeout        time.Duration
-	lifecycleHooks bool
-	rollingRestart bool
-	scope          string
+	client   container.Client
+	notifier *notifications.Notifier
+	c        flags.WatchConfig
 )
 
 var rootCmd = &cobra.Command{
@@ -46,10 +39,11 @@ More information available at https://github.com/containrrr/watchtower/.
 }
 
 func init() {
-	flags.SetDefaults()
 	flags.RegisterDockerFlags(rootCmd)
 	flags.RegisterSystemFlags(rootCmd)
 	flags.RegisterNotificationFlags(rootCmd)
+	flags.SetEnvBindings()
+	flags.BindViperFlags(rootCmd)
 }
 
 // Execute the root func and exit in case of errors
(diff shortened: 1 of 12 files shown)


cmd/root.go
```diff
@@ -74,75 +68,55 @@ func PreRun(cmd *cobra.Command, args []string) {
...
-	schedule, _ := f.GetString("schedule")
-	cronLen := len(schedule)
+	// Update schedule if interval helper is set
+	if viper.IsSet("interval") {
+		if viper.IsSet("schedule") {
+			log.Fatal("only schedule or interval can be defined, not both")
+		}
+		interval := viper.GetInt("interval")
+		viper.Set("schedule", fmt.Sprintf("@every %ds", interval))
+	}
```
piksel: This part does not work correctly, as "interval" has a default value. Should be easy enough to fix though.

piksel: Fixed!

cmd/root.go
```diff
@@ -151,35 +125,33 @@ func Run(c *cobra.Command, names []string) {
...
-	if err := runUpgradesOnSchedule(c, filter); err != nil {
+	if err := runUpgradesOnSchedule(filter); err != nil {
 		log.Error(err)
 	}
 
 	os.Exit(1)
 }
 
-func runUpgradesOnSchedule(c *cobra.Command, filter t.Filter) error {
+func runUpgradesOnSchedule(filter t.Filter) error {
```
simonaronsson, line 148: Any specific reason why you prefer to dig config values out from viper rather than passing the command as in `Run` and `PreRun`?

cmd/root.go
```diff
@@ -221,12 +193,12 @@ func runUpdatesWithNotifications(filter t.Filter) {
 	notifier.StartNotification()
 	updateParams := t.UpdateParams{
```
simonaronsson, line 194: We should probably remove this object altogether and pass the whole `WatchConfig` instead?

simonaronsson (commented):
The pass WatchConfig over fetching from viper question applies to the notification files as well, just didn't want to spam you.

cmd/root.go
piksel: yeah, this was more of a baby steps-thing to keep the LoC down
basically what this was referring to:
> More could be done on that front than what is included in this PR, as I choose to limit this to the intended scope.

cmd/root.go
piksel: the command only has a reference to the flags, but I don't think that was even used before in this, instead it relied on file-global vars. it should probably just be included in the `WatchConfig` struct and have it passed instead

PR #685, opened 2020-11-22
fix host flag

ksurl:
I was following the wiki and trying to use the short format for specifying the docker host with `-h`. It was just spitting out the usage info, and I noticed it was showing `-H`.

The wiki shows the syntax as `--host` or `-h` but `docker run --rm containrrr/watchtower --help` shows `-H`.

### DIFF
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -102,7 +102,7 @@ Environment Variable: NO_COLOR
 Docker daemon socket to connect to. Can be pointed at a remote Docker host by specifying a TCP endpoint as "tcp://hostname:port".
 
 ```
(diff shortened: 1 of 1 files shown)


victorcmoura:
Hi, @ksurl! Thank you for your contribution!

It sounds that the documentation isn't actually matching the Docker Host config implementation: https://github.com/containrrr/watchtower/blob/master/internal/flags/flags.go#L22

LGTM, @simskij

piksel (approved):
Clearly a mistake in the documentation

piksel:
@all-contributors add @ksurl for doc

PR #700, opened 2020-12-07
fix default interval to be the intended value

piksel:
```
time.Hour * 24 / time.Second = 86.4µs
(time.Hour * 24) / time.Second = 86.4µs
(time.Hour * 24).Seconds() = 86400
```
https://play.golang.org/p/2KZUMBe0X_d


simonaronsson:
missed this pr, preferring it over my chaotic hotfix

PR #705, opened 2020-12-09
add defered closer calls for the http clients

simonaronsson:
add defered closer calls for the http client to avoid leaking file descriptors

### DIFF
--- a/pkg/registry/auth/auth.go
+++ b/pkg/registry/auth/auth.go
@@ -37,7 +37,7 @@ func GetToken(container types.Container, registryAuth string) (string, error) {
 	if res, err = client.Do(req); err != nil {
 		return "", err
 	}
-
+	defer res.Body.Close()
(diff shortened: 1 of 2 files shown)


piksel (approved):
This should at the very least apply best practice, and hopefully solve the reported FD leaks:
> The default HTTP client's Transport may not reuse HTTP/1.x "keep-alive" TCP connections if the Body is not read to completion and closed.

It could be the case that the connections are being kept alive for re-use, but never actually reused because of this. But since the `http.Client` doesn't outlive the scope ~~I don't think the connections would actually be reused~~, but that may still be the case due to some static logic in the underlying transports.   
**Edit:** The `DefaultTransport` is static and reused for all `http.Clients`, so that is indeed the case.   

~~Even if that were the case, this might still allow them to be closed when the client is pushed off the stack. The documentation only explicitly mentions that the `Response.Body` needs to be closed.~~

PR #720, opened 2020-12-22
Add WATCHTOWER_INCLUDE_RESTARTING env for include-restarting flag

ilike2burnthing:
Follows up #651 to add environment variable and documentation.

Single letter flag not included as the most obvious option, mirroring `include-stopped` and using `-R`, is already taken by `run-once`.

This is just a c&p job, so no worries if this isn't needed/wanted, it took me all of 2mins.

### DIFF
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -118,6 +118,16 @@ Environment Variable: DOCKER_API_VERSION
              Default: "1.24"
 ```
 
+## Include restarting
+Will also include created and exited containers.
+
+```
+            Argument: --include-restarting
+Environment Variable: WATCHTOWER_INCLUDE_RESTARTING
+                Type: Boolean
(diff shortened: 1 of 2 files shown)


docs/arguments.md
```diff
@@ -118,6 +118,16 @@ Environment Variable: DOCKER_API_VERSION
              Default: "1.24"
 ```
 
+## Include restarting
+Will also include created and exited containers.
```
ilike2burnthing, line 122: Ah damn, my c&p skills failed me, this should be `Will also include restarting containers`, and of course I only see it after it's published 🤦 my bad

@simskij worth another PR or do you wanna just commit this directly?

simonaronsson: I'll fix it. Thanks for letting me know 👍🏼

PR #732, opened 2020-12-26
add details/summary to issue template

piksel:
This will make the log section of an issue expandable (and preformatted) by default.

<details>
<summary>Like this</summary>

```
This is really useful when reading large amounts of text from the console, like logs etc.
The problem is that it's neither-well known, nor easy to get right since you need to put a single 
blank line separating HTML and markdown parts.
```

</details>

Not sure if it's clear enough for people creating the issue how to correctly fill it in. A comment (`<!-- -->`) could be added to explain where to put the log rows, but it might also just be more confusing....

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -28,7 +28,14 @@ If applicable, add screenshots to help explain your problem.
  - Architecture
  - Docker version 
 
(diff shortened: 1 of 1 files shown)


zoispag:
I was totally thinking the same today. I am in favor of adding html comments as well, to instruct what and how.

zoispag (approved):

simonaronsson (approved):

PR #748, opened 2021-01-09
Fix notifications and old instance cleanup

piksel:
- Update `shoutrrr` to v0.3.0
- Add support for discord-via-slack URLs
- Clarify errors relating to old watchtower instance cleanup
- Only treat stopping errors as fatal to support multiple scoped watchtowers using the same image


internal/actions/check.go
```diff
@@ -49,37 +45,23 @@ func cleanupExcessWatchtowers(containers []container.Container, client container
...
 			stopErrors++
 			continue
 		}
 
 		if cleanup {
 			if err := client.RemoveImageByID(c.ImageID()); err != nil {
-				// logging the original here as we're just returning a count
-				logrus.Error(err)
-				cleanupErrors++
+				logrus.WithError(err).Warning("Could not cleanup watchtower, possibly because of other watchtowers instances in other scopes.")
```
simonaronsson:
```suggestion
				logrus.WithError(err).Warning("Could not cleanup watchtower images, possibly because of other watchtowers instances in other scopes.")
```

simonaronsson (approved):

PR #765, opened 2021-01-12
fix: Disallow log level 'trace'

### LINKED ISSUE #759: Setting notification-level to trace disables notifications
This is caused by the first check permitting "trace" as it's a valid logrus level:
https://github.com/containrrr/watchtower/blob/3369d2427c3a7862444f8f34f09aeed3526f1424/pkg/notifications/notifier.go#L21-L26

...but then when grabbing all levels above it at:
https://github.com/containrrr/watchtower/blob/3369d2427c3a7862444f8f34f09aeed3526f1424/pkg/notifications/notifier.go#L27
it uses the `AllLevels` slice from `slackrus`, which doesn't contain `Trace`:
https://github.com/johntdyer/slackrus/blob/f7aae3243a0786c5a974bce71ed951c459876e64/levels.go#L8-L15

...which means that it will return an empty array of accepted levels:
https://github.com/johntdyer/slackrus/blob/f7aae3243a0786c5a974bce71ed951c459876e64/levels.go#L17-L25

zoispag:
Closes #759


piksel (approved):
LGTM

pkg/notifications/notifier.go
```diff
@@ -25,6 +25,10 @@ func NewNotifier(c *cobra.Command) *Notifier {
 	}
 
 	acceptedLogLevels := slackrus.LevelThreshold(logLevel)
+	// slackrus does not allow log level TRACE, even though it's an accepted log level for logrus
+	if len(acceptedLogLevels) == 0 {
+		log.Fatalf("Notifications unsupported log level provided: %s", level)
```
simonaronsson:
```suggestion
		log.Fatalf("Unsupported notification log level provided: %s", level)
```

zoispag:
@simskij is there some permission issue with the runner, that my commit failed and yours not?

simonaronsson:
@zoispag No, I just fixed the workflow. My commit was mainly meant to trigger a rerun and check if it worked 😅

harrytheeskimo:
Please update the Wiki at https://containrrr.dev/watchtower/notifications/ to note that trace is disallowed.
It's still shown as an option.

> --notifications-level (env. WATCHTOWER_NOTIFICATIONS_LEVEL): Controls the log level which is used for the notifications. If omitted, the default log level is info. Possible values are: panic, fatal, error, warn, info, debug or trace.

piksel:
@harrytheeskimo just a FYI; you can directly suggest an edit by clicking the pen icon in the top right of the documentation!

harrytheeskimo:
@piksel thanks, I did try that at the time yesterday, but the page is restricted:

> You need to fork this repository to propose changes.

> Sorry, you’re not able to edit this repository directly—you need to fork it and propose your changes from there instead.

simonaronsson:
Yeah, the flow is that you make an edit in your fork and then create a PR against this repo. We might need to be clearer about that. Sorry if it tripped you up.

PR #766, opened 2021-01-12
chore(ci): disable fail-fast for pr tests

piksel:
This will make the `Test` step continue even if one platform errors out (not naming names here, wintendo). Also useful to get an indication of whether the problem is OS-specific when all builds should work normally.


zoispag (approved):
I was about to open the same PR!

PR #793, opened 2021-01-21
Pre-update lifecycle hook

### LINKED ISSUE #649: Make watchtower skip update if pre-update lifecycle hook exits with a non-zero exit code

yrien30:
Will close #649


yrien30:
Not sure why the build failed. Anyone know?

kevinlul:
How can I help out? 👀 Just testing?

yrien30:
> How can I help out? 👀 Just testing?

Great if you could test it and maybe come whit suggestion on improving the documentation. I'm also not sure if there should be a own flag for skipping the lifecycle hooks when the container is stoped or restarting.

yrien30:
I have uploaded it to ghcr.io. So If anyone would test it then you will find it [here](https://github.com/users/yrien30/packages/container/package/watchtower) Its build for arm64.

yrien30:
I have now tested that this works.

Exit code 1:
image
Exit code 75:
image
Container that needs update and is not running. Previously it would not get updated because the preupdate script can only run when the container is running..
image
Exit code 0: 
image

yrien30:
Making this a draft again since I found out that the container is not created after it had been deleted. I need figure out why..

image

yrien30:
Tested that the container gets started again after latest fix:
image

simonaronsson:
Awesome! Thanks for your contribution! I can't review it right now, but will get back to it as soon as possible.

pkg/container/client.go
```diff
@@ -382,15 +382,15 @@ func (client dockerClient) ExecuteCommand(containerID string, command string, ti
...
-		return err
+		return true,err 
 	}
 
-	return nil
+	return skipUpdate,nil
 }
 
-func (client dockerClient) waitForExecOrTimeout(bg context.Context, ID string, execOutput string, timeout int) error {
+func (client dockerClient) waitForExecOrTimeout(bg context.Context, ID string, execOutput string, timeout int) (bool,error) {
```
piksel: Even though it's called just above, I really think bools should always be named

```suggestion
func (client dockerClient) waitForExecOrTimeout(bg context.Context, ID string, execOutput string, timeout int) (skipUpdate bool, err error) {
```

pkg/container/client.go
```diff
@@ -420,13 +420,17 @@ func (client dockerClient) waitForExecOrTimeout(bg context.Context, ID string, e
 		if len(execOutput) > 0 {
 			log.Infof("Command output:\n%v", execOutput)
 		}
-		if execInspect.ExitCode > 0 {
-			log.Errorf("Command exited with code %v.", execInspect.ExitCode)
-			log.Error(execOutput)
+		
+		if execInspect.ExitCode == 75{
```
piksel: Perhaps use a const?

piksel (commented):
This does look good in general, but there really ought to be tests for the functionality.

Also, some run `go fmt` to fix things like spaces between arguments.

yrien30:
> This does look good in general, but there really ought to be tests for the functionality.
> 
> Also, some run `go fmt` to fix things like spaces between arguments.

I have added some test and run `go fmt`. Have also retested everything like I did above manualy after I merged the changes form main.

yrien30:
Any chance to get an update on whether this could be merged or need more work?

subdavis:
Without tests, this is a 100 LoC change that several people clearly care about.  Thanks to @yrien30 for putting this PR together.

As a show of good faith, I've started sponsoring https://github.com/sponsors/simskij

If this could be reviewed, merged, and released sometime soon, that would be greatly appreciated.

simonaronsson (approved):

simonaronsson:
LGTM! Thanks for the contribution! Let's try this in `:latest-dev` for a bit to make sure it's stable.

PR #802, opened 2021-01-27
Log schedule changed to info level

### LINKED ISSUE #801: Scheduler next run log as debug level instead of info
**Describe the bug**

When running watchover with the argument `--schedule, -s` the scheduled next run log is under `DEBUG` level. 

It should be as `INFO` level as the end user would not have any clue of the correct execution of the scheduled action if debug mode is not enabled.

`time="2021-01-27T05:33:05+01:00" level=debug msg="Scheduled next run: 2021-01-28 05:33:00 +0100 CET"
`

Otherwise, **an `INFO` log when the cron is started and finished would be as usefull**. But at the moment (without using the --debug option) there is no way to know if watchtower is working as intended.

**Setup using Docker Compose**

```
version: "3"

services:
  watchtower:
    image: containrrr/watchtower:latest
    container_name: watchtower
    restart: unless-stopped
    environment:
      TZ: "Europe/Madrid"
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    command: --debug --cleanup --schedule "0 0 5 * * *"
    #command: --cleanup --schedule "0 0 5 * * *"
    logging:
      driver: "json-file"
      options:
        max-size: "5m"
        max-file: "10"
```

fizcris:
Closes #801

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -209,7 +209,7 @@ func runUpgradesOnSchedule(c *cobra.Command, filter t.Filter) error {
 
 			nextRuns := cron.Entries()
 			if len(nextRuns) > 0 {
-				log.Debug("Scheduled next run: " + nextRuns[0].Next.String())
+				log.Info("Scheduled next run: " + nextRuns[0].Next.String())
 			}
 		})
 


simonaronsson:
Thank you so much for your contribution! 🙏🏼 

As stated in https://github.com/containrrr/watchtower/issues/801#issuecomment-770222413, this would lead to a lot of notification spam. I think having to add the debug flag to verify something and then turn it off again is preferable to getting one additional notification for every cycle. We might reconsider this in the future, but for now, it will stay as it is.

PR #809, opened 2021-01-30
feat: include additional info in startup

piksel:
I am investigating alternative startup messages for Watchtower, based on the common issues users seem to have. This is the proposed output from run once:
```
$ ./watchtower --run-once --notifications gotify,shoutrrr --scope 15xPMII foo bar --debug true --label-enable
INFO[0000] Watchtower v1.1.6
Using notifications: gotify, smtp
Only checking containers with name "foo" or "bar" or "true", using enable label, in scope "15xPMII"
Running a one time update.
Waiting for the notification goroutine to finish
Waiting for the notification goroutine to finish
```
And this is the corresponding one for interval/schedule/--not-run-once:
```
./watchtower --notifications gotify,shoutrrr foo bar --label-enable
INFO[0001] Watchtower v1.1.6
Using notifications: gotify, smtp
Only checking containers with name "foo" or "bar", using enable label
Scheduling first run: 2021-01-31 13:53:08 +0100 CET
Note that the first check will be performed in 23 hours, 59 minutes, 59 seconds
INFO[0039] Waiting for running update to be finished...
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+VERSION=$(git describe)
+echo "Building $VERSION..."
+go build -o watchtower -ldflags "-X github.com/containrrr/watchtower/cmd.version=$VERSION"
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -38,7 +38,8 @@ var (
 	lifecycleHooks bool
 	rollingRestart bool
 	scope          string
-	version        string = "v0.0.0-unknown"
+	// Set on build using ldflags
+	version = "v0.0.0-unknown"
 )
 
 var rootCmd = NewRootCommand()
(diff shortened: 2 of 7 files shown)


simonaronsson:
Anything more you'd like to do here, @piksel, or is it ready for review?

piksel:
Well, it's currently outputting what was presented in the team discussion (I'll update the PR with discussion message).
It's marked as a draft because it's in more of a "suggestion" state (one might call it a... draft 😁), since I'm not sure what we actually want to display/notify on startup. 
So it's ready for review, but not of code quality, but more of functionality, if that makes sense...

simonaronsson:
> Well, it's currently outputting what was presented in the team discussion (I'll update the PR with discussion message).
> It's marked as a draft because it's in more of a "suggestion" state (one might call it a... draft 😁), since I'm not sure what we actually want to display/notify on startup.
> So it's ready for review, but not of code quality, but more of functionality, if that makes sense...

From a functional perspective, I'm satisfied with using what you suggest as a first released iteration. 👍🏼

PR #827, opened 2021-02-17
Update HTTP API docs

dapplion:
Docs don't match source


benben (approved):
I tested that and it works, please merge :)

victorcmoura (approved):
Hi, @dapplion! Thanks for your contribution. I believe we just have to wait for @simskij to double-check and merge it.

Issue #873, opened 2021-03-31
Gotify is sending one single message with all notifications from watchtower


cerealconyogurt:
**Describe the bug**
Before updating from v1.1.5 to v.1.2.1 watchtower was sending each notification to gotify separately, so I was receiving a message from gotify for each message from watchtower. See below an example

![image](https://user-images.githubusercontent.com/49502224/113161217-8d25d480-9214-11eb-9da1-8a26a17ee8f0.png)

but now the notifications coming in just one message:

![image](https://user-images.githubusercontent.com/49502224/113162500-bc891100-9215-11eb-9cae-9e1cd57e23ea.png)

**Expected behavior**
I am expecting to see one message per one notification. It is hard to read which docker has been updated when all notifications are in one message in gotify

**Environment**
 - Platform Asustor AS5202T
 - Linux x86_64 (kernel 4.14.0.x)
 - Docker version 19.03.8 (API: 1.40)

**log**
`today at 1:31 AM  time="2021-03-31T01:31:15-03:00" level=info msg="Found new haugene/transmission-openvpn:latest image (sha256:26319b02276cce1fd61189fd97cafc830cb35b128375f77b7c126f4756f4e532)"`
`today at 1:31 AM  time="2021-03-31T01:31:42-03:00" level=info msg="Found new quay.io/invidious/invidious:latest image (sha256:1f88fd91398fbfe85dd871bbd7ae4bfe4823adca912e710b781c55caf5753cbf)"`
`today at 1:31 AM  time="2021-03-31T01:31:50-03:00" level=info msg="Found new excalidraw/excalidraw:latest image (sha256:2a0dc9a84a8272835837fda3e7ced670b2d04e44a3b64a8759a796d831514ae5)"`
`today at 1:32 AM  time="2021-03-31T01:32:09-03:00" level=info msg="Found new amir20/dozzle:latest image (sha256:9cf27bca912f322057abab409a0f0f243c5907805047657a4ec42cac5abe459d)"`

scottjl:
this is the same format shoutrrr uses and i too wouldn't mind separate messages for each container vs. one very long message.

piksel:
The reason for this change was that all messages belonging to a single run would be "connected", so that if you had 20 containers to check, you wouldn't get 20 notifications (minimum!) for every run.
We could perhaps add an option to split the messages if this is really important, but I would suggest exploring other alternatives to not increase the maintenance/testing/development burden.

[...]

PR #888, opened 2021-04-08
feat(log): use short image/container IDs in logs

piksel:
This should clean up the logging output and only refer to images and containers using their short hashes.

ref #873

### DIFF
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -147,8 +147,10 @@ func (client dockerClient) StopContainer(c Container, timeout time.Duration) err
 		signal = defaultStopSignal
 	}
 
+	shortID := ShortID(c.ID())
+
 	if c.IsRunning() {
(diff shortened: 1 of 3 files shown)


pkg/container/client.go
```diff
@@ -284,13 +286,13 @@ func (client dockerClient) PullImage(ctx context.Context, container Container) e
...
-	if opts.RegistryAuth != "" {
-		log.Debug("Credentials loaded")
-	}
 	if err != nil {
 		log.Debugf("Error loading authentication credentials %s", err)
 		return err
 	}
+	if opts.RegistryAuth != "" {
+		log.Debug("Credentials loaded")
+	}
```
piksel, line 295: Not actually part of the intended changes, but got a lint warning about the use of opts before checking if an error was returned. Shouldn't matter, but doing it in this order is at least more correct.

simonaronsson (approved):

Issue #892, opened 2021-04-09
HEAD request log messages


simonaronsson:
> Hi!
> 
> No, that is by design. Since DockerHub started to enforce rate limiting, we added support for doing HEAD requests to compare digests instead of pulling the whole manifest. We decided to put that as a warning as watchtower no longer doing head requests would mean their usage quota will be consumed.
>
_Originally posted by @simskij in https://github.com/containrrr/watchtower/discussions/890#discussioncomment-590012_

A lot of users have reported that this message is annoying or confusing, so I'd like to arrange some kind of vote about how to proceed. Personally, I'd much rather have watchtower warn about head request errors than not.

What do y'all feel? Let's do a vote using emojis. 

🎉  - Move warning to debug level
🚀  - Add an option to suppress the warning (ie. something like `--no-head-warnings`)
👀  - Keep as is.

educatedCaveman:
I'd be fine with either moving it to debug or adding an option to suppress

Dulanic:
I'd recommend exclusion being explicit not implicit. I just fixed mine because I saw so many of those messages and realized I had network mode none which was causing it. People should research it, but have the ability to easily turn it off. If it's implicit, I would not have known.

piksel:
Thanks @Dulanic, it's much harder to justify being annoying 😅.

simonaronsson:
> I'd recommend exclusion being explicit not implicit. I just fixed mine because I saw so many of those messages and realized I had network mode none which was causing it. People should research it, but have the ability to easily turn it off. If it's implicit, I would not have known.

That's great feedback! Thanks!

simonaronsson:
We've decided to go for the explicit opt-out route, making the logs default for ghcr and dockerhub, but allowing users to turn them on for **all** repos if they wish, or off for **all** repos. Thank you for your input!

PR #912, opened 2021-04-20
feat: make head pull failure warning toggleable

piksel:
A suggestion for resolving #892.

This will add the command line argument `--warn-on-head-failure`, which can be set to `always`, `auto` (default) and `never`.
Where `auto` is the current behaviour, where registries that are known to work (and especially docker hub) will yield a warning when HEAD requests fail.
Setting it to `never` disables the warning, and `always` warns regardless of whether the registry is known to support HEAD. The last one is mostly useful for testing is not expected to be used much for the average user.

It also changes what the warnings contain. They now always include the image name and the reason for the HEAD failure. 
The disabled warnings are still logged as `debug`, so you can still find them in your logs if you have `--debug` enabled.


simonaronsson (approved):

piksel (commented):
Needs a `go fmt`, there's some mixed indentations

PR #916, opened 2021-04-21
Allow running periodic updates with enabled HTTP API

### LINKED ISSUE #897: Upgrades on schedule not working when API enabled
**To Reproduce**
Enable http api: 
-e WATCHTOWER_HTTP_API_UPDATE=true -e WATCHTOWER_HTTP_API_TOKEN=Token

**Expected behavior**
I expect watchtower to check for updates every 10 minues since I have enabled this setting:  -e WATCHTOWER_POLL_INTERVAL=600

No updates since 2021-04-07T11:31:34Z time now is 2021-04-13T10:58:34Z
image

I suspect that api server is blocking for update scheduling from starting:
https://github.com/containrrr/watchtower/blob/aa3ed8833f484b3c016f8c76260442b1487a91c4/cmd/root.go#L182-L186

**Environment**
 - arm64

DasSkelett:
Edit by @simskij: closes #897 

## Motivation
I'd like to run Watchtower both with infrequent periodic updates for images I don't build myself, and have the option to quickly redeploy custom images when they're rebuilt using the HTTP API.
Currently this is not possible, because enabling the HTTP Update API disables scheduled runs.

In theory this would also be possible using multiple Watchtower instances with two scopes, one for "public" images and one for self-built images; however now you need to deal with labels and make sure every container has the correct scope label, which adds unwanted complexity to the setup.
Alternatively you could set up a Cron job to curl `/v1/update` periodically, but this also gets ugly fast given that we're in Docker, and it also splits the "configuration" of Watchtower into two locations. Also the primary purpose for Watchtower is to replace hacky cron jobs to update Docker containers.

Instead, it would be very handy if Watchtower allows running periodic updates and the HTTP API for updates at the same time.

## Changes
Add a new `--http-api-periodic-polls` option with the `WATCHTOWER_HTTP_API_PERIODIC_POLLS` environment variable, defaulting to false.
It has only an effect if the HTTP API mode (`--http-api-update`) is enabled.
If `--http-api-periodic-polls` is passed, the HTTP API doesn't block, and thus allows periodic polls as well (with whatever `--interval` or `--schedule` is given, defaulting to every 24 hours).

I'm open for suggestions to rename this option if you've got something better in mind. I'm not entirely happy with the word "periodic" since it isn't used anywhere else so far. I've considered "scheduled", but it might mislead to think it only affects the `--schedule` option and not `--interval`.

The HTTP API and `runUpgradesOnSchedule()` had separate locks to prevent multiple updates running at the same time.
Since there's the possibility that the schedule triggers while an HTTP API poll is running, or the other way around, I needed to add a global lock. The channel is passed to `runUpgradesOnSchedule()` and `api.update.New()`, which now use this channel for locking instead of making their own.
Both functions still create a new channel if the argument is `nil`, which might simplify future unit testing or other changes.

I've also extended the docs accordingly. If I should clarify or reword something, please do tell.

## Further considerations
I also thought about enabling this behaviour by default (inverting the option and renaming it to`--http-api-only`), since this might be what most people expect, but this could be a "breaking" change, since it differs from the current behaviour.
But if you say it's worth it, I would be happy to do that.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -156,6 +156,7 @@ func Run(c *cobra.Command, names []string) {
 	runOnce, _ := c.PersistentFlags().GetBool("run-once")
 	enableUpdateAPI, _ := c.PersistentFlags().GetBool("http-api-update")
 	enableMetricsAPI, _ := c.PersistentFlags().GetBool("http-api-metrics")
+	unblockHTTPAPI, _ := c.PersistentFlags().GetBool("http-api-periodic-polls")
 	apiToken, _ := c.PersistentFlags().GetString("http-api-token")
 
 	if rollingRestart && monitorOnly {
@@ -180,10 +181,14 @@ func Run(c *cobra.Command, names []string) {
 		logNotifyExit(err)
 	}
 
+	// The lock is shared between the scheduler and the HTTP API. It only allows one update to run at a time.
+	updateLock := make(chan bool, 1)
+	updateLock <- true
+
 	httpAPI := api.New(apiToken)
 
 	if enableUpdateAPI {
-		updateHandler := update.New(func() { runUpdatesWithNotifications(filter) })
+		updateHandler := update.New(func() { runUpdatesWithNotifications(filter) }, updateLock)
 		httpAPI.RegisterFunc(updateHandler.Path, updateHandler.Handle)
 	}
 
@@ -192,11 +197,11 @@ func Run(c *cobra.Command, names []string) {
 		httpAPI.RegisterHandler(metricsHandler.Path, metricsHandler.Handle)
 	}
 
-	if err := httpAPI.Start(enableUpdateAPI); err != nil {
+	if err := httpAPI.Start(enableUpdateAPI && !unblockHTTPAPI); err != nil {
 		log.Error("failed to start API", err)
 	}
 
-	if err := runUpgradesOnSchedule(c, filter, filterDesc); err != nil {
+	if err := runUpgradesOnSchedule(c, filter, filterDesc, updateLock); err != nil {
 		log.Error(err)
 	}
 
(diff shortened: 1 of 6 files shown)


DasSkelett:
I just noticed that the locks are all local and not shared between HTTP API and the scheduler, so I'll take a stab at trying to create a global lock we can pass to both to avoid running scheduled and API-triggered polls at the same time.

Also looking at the logs of the `ubuntu-latest` test I'm unsure why the test fails and if I caused it. A race between creating the server and feeding the metrics?

simonaronsson:
> Also looking at the logs of the `ubuntu-latest` test I'm unsure why the test fails. [...] A race between creating the server and feeding the metrics?

Yeah. This is fixed in `feat/head-failure-toggle`, however, we still need to improve the code coverage of that branch a bit to be able to merge it with `main`. Great contribution! I still haven't reviewed the code, but the concept definitely makes sense!

piksel:
Just a thought, wouldn't manually specifying a schedule or interval warrant the enabling of the periodic updates as well?
It would be breaking behaviour, but it seems like a more obvious way to inerpret the command line/environment args.

It should probably also write a startup message when running in "manual" mode, with `schedMessage` being something like:
`Waiting for manual dispatch`, but that is probably outside the scope of this PR...

simonaronsson:
> Just a thought, wouldn't manually specifying a schedule or interval warrant the enabling of the periodic updates as well?
> It would be breaking behaviour, but it seems like a more obvious way to inerpret the command line/environment args.
> 
> It should probably also write a startup message when running in "manual" mode, with `schedMessage` being something like:
> `Waiting for manual dispatch`, but that is probably outside the scope of this PR...

Your thought is definitely valid and warranted. I'll go ahead and fork it off into a separate issue though.

piksel (approved):

simonaronsson (approved):
Agree that this should have tests, but lets get this into the code base. @DasSkelett, if you've got time to add some tests, that would be greatly appreciated, but we can keep that as a separate PR in that case. 

Thank you for your great contribution! 🎉

simonaronsson:
@all-contributors add @DasSkelett for code

DasSkelett:
Testing `Run()` for the correct behaviour won't be that easy, since there is nothing set up for `root.go` currently, and I'm also not sure how I could look at what's happening inside, but I'll have a think and see if I can come up with something. Maybe the test framework offers some tricks.

But I could copy the relevant parts from `Run()`, i.e. create a shared lock and make sure one trigger doesn't run while the other one is active.

PR #930, opened 2021-04-24
test: fix metrics api test stability

piksel:
Should permanently solve the intermittent test failures.


simonaronsson:
I know we found out that the actual reason for the problems was how we do the calls, nevertheless I think the httptest part of this PR is a welcome change as well as it makes the tests less fragile overall. Good job!

simonaronsson (approved):

PR #981, opened 2021-06-06
Session report collection and report templates

piksel:
PoC for collecting a report containing session stats. Updated to not replace the current notification templates, but instead make the report-based notifications an opt-in (using the `--notification-report` flag)

---

This will generate the following report for each session:
https://github.com/containrrr/watchtower/blob/c34de89cc075fba0e01be7888bc7a72136393df4/pkg/types/report.go#L3-L10
where each `ContainerReport` contains
https://github.com/containrrr/watchtower/blob/c34de89cc075fba0e01be7888bc7a72136393df4/pkg/types/report.go#L12-L20

This report is passed to the template together with the log entries (the old template data):
https://github.com/containrrr/watchtower/blob/c34de89cc075fba0e01be7888bc7a72136393df4/pkg/notifications/shoutrrr.go#L206-L209

This makes it possible to send notifications like these instead of the full logs:
```
3 Scanned, 1 Updated
 - /baz (alpine:latest): Updated
 - /bar (nginx:latest): Fresh
 - /bar (redis:latest): Stale
Found new alpine:latest image (5de788243aca)
Stopping /baz (a9faebf599b0) with SIGTERM
Creating /baz
```

...using template:
```
{{with .Report -}}
  {{len .Scanned}} Scanned, {{len .Updated}} Updated
  {{range .Scanned}}
     - {{.Name}} ({{.ImageName}}): {{.State}}
  {{end}}
{{- end}}
{{range .Entries}}
  {{- .Message -}}
{{end}}
```

*Note that the above still contains the old "log" notifications at the bottom as well, but the point is that you don't have to include them.*

<details>
<summary>Report container classifications:</summary>

image

</details>


pkg/container/interface.go.go
```diff
@@ -0,0 +1,9 @@
+package container
+
+// Interface is the minimum common Container interface
+type Interface interface {
+	ID() string
+	Name() string
+	SafeImageID() string
+	ImageName() string
+}
```
piksel: This could be replaced with an addition to `types.Container` interface instead.

simonaronsson (approved):

simonaronsson (approved):

PR #994, opened 2021-06-18
feat: allow hostname override for notifiers

nightah:
As it currently stands all notifiers utilise `os.Hostname` to populate their titles/subjects.

When utilising Docker with a bridged network if you set the hostname for a container to an external DNS hostname Docker's internal DNS resolver will override said hostname for all containers within the bridged network.

This change allows a user to specify what hostname should be represented in the notifications without having to change the `os.Hostname`.

### DIFF
--- a/docs/notifications.md
+++ b/docs/notifications.md
@@ -42,6 +42,7 @@ To receive notifications by email, the following command-line options, or their 
 - `--notification-email-server-password` (env. `WATCHTOWER_NOTIFICATION_EMAIL_SERVER_PASSWORD`): The password to authenticate with the SMTP server with. Can also reference a file, in which case the contents of the file are used.
 - `--notification-email-delay` (env. `WATCHTOWER_NOTIFICATION_EMAIL_DELAY`): Delay before sending notifications expressed in seconds.
 - `--notification-email-subjecttag` (env. `WATCHTOWER_NOTIFICATION_EMAIL_SUBJECTTAG`): Prefix to include in the subject tag. Useful when running multiple watchtowers.
+- `--notification-email-subjecthostname` (env. `WATCHTOWER_NOTIFICATION_EMAIL_SUBJECTHOSTNAME`): Custom hostname specified in subject. Useful to override the hostname.
 
 Example:
 
--- a/docs/notifications.md
+++ b/docs/notifications.md
@@ -25,6 +25,7 @@ comma-separated list of values to the `--notifications` option
 ## Settings
 
 - `--notifications-level` (env. `WATCHTOWER_NOTIFICATIONS_LEVEL`): Controls the log level which is used for the notifications. If omitted, the default log level is `info`. Possible values are: `panic`, `fatal`, `error`, `warn`, `info`, `debug` or `trace`.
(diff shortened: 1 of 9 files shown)


internal/flags/flags.go
```diff
@@ -240,6 +240,12 @@ Should only be used for testing.`)
 		viper.GetString("WATCHTOWER_NOTIFICATION_EMAIL_SUBJECTTAG"),
 		"Subject prefix tag for notifications via mail")
 
+	flags.StringP(
+		"notification-email-subjecthostname",
+		"",
+		viper.GetString("WATCHTOWER_NOTIFICATION_EMAIL_SUBJECTHOSTNAME"),
+		"Subject custom hostname for notifications via mail")
+
```
```suggestion
	flags.StringP(
		"notification-hostname",
		"",
		viper.GetString("WATCHTOWER_NOTIFICATION_HOSTNAME"),
		"Custom hostname for notification titles")
```

piksel (changes requested):
The title isn't tied to email anymore (hence why it's in `pkg/notifications/notifier.go`). It should not be too difficult to set the notification hostname either to the flag value, or if empty, `os.Hostname`.

There are more places to change than the one below obviously though.

internal/flags/flags.go
nightah: Adjusted in https://github.com/containrrr/watchtower/pull/994/commits/b3bbc3ca0c676c04e700aff5e337cd5ab005b48a.

Also added the ability to override the hostname for all notifiers.

docs/notifications.md
```diff
@@ -42,6 +42,7 @@ To receive notifications by email, the following command-line options, or their
 - `--notification-email-server-password` (env. `WATCHTOWER_NOTIFICATION_EMAIL_SERVER_PASSWORD`): The password to authenticate with the SMTP server with. Can also reference a file, in which case the contents of the file are used.
 - `--notification-email-delay` (env. `WATCHTOWER_NOTIFICATION_EMAIL_DELAY`): Delay before sending notifications expressed in seconds.
 - `--notification-email-subjecttag` (env. `WATCHTOWER_NOTIFICATION_EMAIL_SUBJECTTAG`): Prefix to include in the subject tag. Useful when running multiple watchtowers.
+- `--notification-email-subjecthostname` (env. `WATCHTOWER_NOTIFICATION_EMAIL_SUBJECTHOSTNAME`): Custom hostname specified in subject. Useful to override the hostname.
```
piksel: Documentation still reflects the previous flag. It should probably be moved to `## Settings` as well
```suggestion
- `--notifications-hostname` (env. `WATCHTOWER_NOTIFICATIONS_HOSTNAME`): Custom hostname for notification titles
```

piksel (changes requested):
That was fast!

The way it's implemented here is probably the simplest (in a good way). A larger overhaul of the notification flag handling should probably be done at a later stage anyway.

Just one comment:

docs/notifications.md
nightah: Sorry I did just notice this afterwards, fixed though my wording isn't exactly the same, let me know if you would like me to adjust it.

piksel: Well, never mind then :D

piksel (approved):

PR #1010, opened 2021-07-01
fix(notifications): default templates and logic

### LINKED ISSUE #1008: Discord/Slack notifications being sent for runs with no updates
**Describe the bug**

On the latest dev image and for the last week, I noticed that Watchtower is now sending notifications for every single run, even if there are no updates. Previously, it would only send a notification if a container was actually updated. This is creating even more notification spam than usual (I was hoping Watchtower would go the other way where notifications became limited to summaries of runs where containers were updated).

**To Reproduce**

1. Use latest dev version of Watchtower and configure Discord notifications. Not sure if this issue exists for Slack or Shoutrr as well, but I am using Discord via the Slack configuration variables (which works outside this issue).

```yaml
  watchtower:
    image: containrrr/watchtower:latest-dev
    container_name: watchtower
    environment:
      - TZ=${TZ}
      - WATCHTOWER_CLEANUP=true
      - WATCHTOWER_REMOVE_VOLUMES=true
      - WATCHTOWER_SCHEDULE=0 0 * * * *
      - WATCHTOWER_INCLUDE_STOPPED=true
      - WATCHTOWER_INCLUDE_RESTARTING=true
      - WATCHTOWER_NOTIFICATIONS=slack
      - WATCHTOWER_NOTIFICATION_SLACK_HOOK_URL=https://discord.com/api/webhooks/${WATCHTOWERWEBHOOK}/slack
      - WATCHTOWER_NOTIFICATION_SLACK_IDENTIFIER=Watchtower
      - "WATCHTOWER_NOTIFICATION_SLACK_ICON_EMOJI=:whale:"
      - WATCHTOWER_NOTIFICATION_SLACK_ICON_URL=https://raw.githubusercontent.com/containrrr/watchtower/master/logo.png
      - DOCKER_CONFIG=/config
    volumes:
      - ${CONFDIR}/watchtower:/config
      - /var/run/docker.sock:/var/run/docker.sock
    hostname: ${HOSTNAME}
    logging: *default-logging
    labels: *default-labels
    restart: unless-stopped
```
2. Wait for a run without any updates, check notifications    

**Expected behavior**

Notifications are only sent if a container is being updated on a run

**Screenshots**

image
**Environment**

`latest-dev` tag, currently running version `v1.3.0-19-gea300b7`

piksel:
Follow up/continuation of #981.

This PR will fix the default report template to only output something if `Updated` or `Failed` is non-zero. It also fixes the non-session notifications (like the startup info) which was just broken using the report notifications.
Note that this aligns the notification behaviour roughly with how it works using legacy notifications, as no notifications will be sent when everything is "normal"/already updated.

<details>

<summary><strong>the new default report template</strong></summary>

```go
{{- if .Report -}}
  {{- with .Report -}}
    {{- if ( or .Updated .Failed ) -}}
{{len .Scanned}} Scanned, {{len .Updated}} Updated, {{len .Failed}} Failed
      {{- range .Updated}}
- {{.Name}} ({{.ImageName}}): {{.CurrentImageID.ShortID}} updated to {{.LatestImageID.ShortID}}
      {{- end -}}
      {{- range .Fresh}}
- {{.Name}} ({{.ImageName}}): {{.State}}
      {{- end -}}
      {{- range .Skipped}}
- {{.Name}} ({{.ImageName}}): {{.State}}: {{.Error}}
      {{- end -}}
      {{- range .Failed}}
- {{.Name}} ({{.ImageName}}): {{.State}}: {{.Error}}
      {{- end -}}
    {{- end -}}
  {{- end -}}
{{- else -}}
  {{range .Entries -}}{{.Message}}{{"\n"}}{{- end -}}
{{end -}}
```

</details>

To prevent the notifications from being sent, it skips notifications whenever the output of the template (whether legacy or report templates) is an empty string.
This fixes #1008 

It also adds a "special" data field for logging that just marks a message as non-notification. This is so that we can still log using the appropriate log levels without having to worry about it getting sent as notifications.
This mainly concerns logs about the notifications themselves, which would cause loops etc. 
Any log messages with the field `notify` set to `no` is ignored.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"github.com/containrrr/watchtower/internal/meta"
 	"math"
+	"net/http"
 	"os"
 	"os/signal"
 	"strconv"
@@ -197,7 +198,7 @@ func Run(c *cobra.Command, names []string) {
 		httpAPI.RegisterHandler(metricsHandler.Path, metricsHandler.Handle)
 	}
 
-	if err := httpAPI.Start(enableUpdateAPI && !unblockHTTPAPI); err != nil {
+	if err := httpAPI.Start(enableUpdateAPI && !unblockHTTPAPI); err != nil && err != http.ErrServerClosed {
 		log.Error("failed to start API", err)
 	}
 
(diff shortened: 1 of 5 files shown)


piksel:
Also, this enables having the startup message being *logged* even if it's not wanted as a notification, it also changes said startup message to be a series of messages instead of a single one with newlines in it (which was not really supported in logfmt):
```
INFO[0001] Watchtower v0.0.0-unknown
INFO[0001] Using notifications: telegram
INFO[0001] Checking all containers (except explicitly disabled with label)
INFO[0001] Running a one time update.
```
They are still batched up and sent as a single notification, so there should be no difference in the notifications.

Also, this is how the closing messages look, no more raw printing to stdout:
```
DEBU[0002] No new images found for /testcontainer1
DEBU[0002] Session done: 1 scanned, 0 updated, 0 failed
INFO[0002] Waiting for the notification goroutine to finish  notify=no
INFO[0002] Skipping notification due to empty message    notify=no
```
And the same goes for notification errors when sending:
```
ERRO[0002] Failed to send shoutrrr notification          error="failed to send notification to \"xxxxxx\", response status code 401 Unauthorized" index=0 notify=no service=telegram
```
...and for template issues:
```
FATA[0001] Notification template error                   error="template: :1:3: executing \"\" at <.InvalidProp>: can't evaluate field InvalidProp in type notifications.Data" notify=no
```

ilike2burnthing:
Is the Docker log supposed to look like this when no notifications are configured?
```
time="2021-09-21T18:46:22Z" level=info msg="Skipping notification due to empty message" notify=no
time="2021-09-21T18:47:21Z" level=info msg="Skipping notification due to empty message" notify=no
time="2021-09-21T18:48:21Z" level=info msg="Skipping notification due to empty message" notify=no
```
(changed the poll interval to 60 to test if it was related to the timing of the log entries)

piksel:
Sort of. It's not really necessary when you don't have any notifications of course...

ilike2burnthing:
Any way to disable it?

PR #1029, opened 2021-07-22
Prefer long flags in quick start example

rootulp:
`--detach` is clearer (to a new Docker user) than `-d`. Similarly `--volume` is clearer than `-v`.

Motivation: https://changelog.com/posts/use-long-flags-when-scripting
Referred to: https://docs.docker.com/engine/reference/commandline/run/

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,9 @@
 With watchtower you can update the running version of your containerized app simply by pushing a new image to the Docker Hub or your own image registry. Watchtower will pull down your new image, gracefully shut down your existing container and restart it with the same options that were used when it was deployed initially. Run the watchtower container with the following command:
 
 ```
-$ docker run -d \
+$ docker run --detach \
     --name watchtower \
-    -v /var/run/docker.sock:/var/run/docker.sock \
+    --volume /var/run/docker.sock:/var/run/docker.sock \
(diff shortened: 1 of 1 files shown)


piksel (approved):
Agreed. It's clearer what every argument is doing, and if you already know what they do you probably also know the short hands (or you look them up if you care enough about make your commands shorter 😁)

zoispag:
I disagree here. All the docker examples I was finding and following on the internet when I was learning docker were **ALWAYS** using `-d`, `-p`, `-v`, `-e`, etc. 
Even now, experienced in Docker, `--detach` reads funny and is unusual to see it, imho.

piksel:
@zoispag Indeed. That's probably why they were used in the docs in the first place. I think the main problem is with people who just copy commands from the docs and try to modify them to suite their needs. 
The whole command can be fairly opaque to someone who don't know what flags take an argument etc.
Usually, `-v` maps to `--verbose` for example, which doesn't take an argument. 🤷‍♀️

rootulp:
> Usually, -v maps to --verbose for example, which doesn't take an argument. 🤷‍♀️

Agreed. I was similarly confused by this naming conflict. I think the longer format flags remove the risk of confusion.

PR #1061, opened 2021-08-31
update dependencies (sane go.mod)

piksel:
this should both simplify the dependencies (and the go.mod), bring us closer to the current docker client and remove dependency on the old `runc` (which is flagged as vulnerable by dependabot, even though the code is not used in watchtower itself).


pkg/registry/trust.go
```diff
@@ -96,6 +97,10 @@ func CredentialsStore(configFile configfile.ConfigFile) credentials.Store {
 }
 
 // EncodeAuth Base64 encode an AuthConfig struct for transmission over HTTP
-func EncodeAuth(auth types.AuthConfig) (string, error) {
-	return command.EncodeAuthToBase64(auth)
+func EncodeAuth(authConfig types.AuthConfig) (string, error) {
+	buf, err := json.Marshal(authConfig)
+	if err != nil {
+		return "", err
```
piksel, line 103: This basically cannot happen, since the passed struct is always serializable (which also makes it impossible to write a test for it). Still, it wouldn't be right to ignore the error, since something might change that makes it possible (or cosmic radiation etc)  🤷‍♀️

simonaronsson (approved):

Issue #939, opened 2021-04-26
API changes for selecting whether to use HTTP API triggers, scheduled triggers, or both


simonaronsson:
> Just a thought, wouldn't manually specifying a schedule or interval warrant the enabling of the periodic updates as well?
> It would be breaking behaviour, but it seems like a more obvious way to inerpret the command line/environment args.
>
> It should probably also write a startup message when running in "manual" mode, with `schedMessage` being something like:
> `Waiting for manual dispatch`, but that is probably outside the scope of this PR...

_Originally posted by @piksel in https://github.com/containrrr/watchtower/issues/916#issuecomment-827035360_

simonaronsson:
This is an interesting observation that I think requires that we put some more thought into how the user APIs are currently organized. One issue currently is that we really can't *solely* use the existence of both as an indicator of whether to enable both triggers or not, as `--interval` has a "sensible default" set.

However, I would like to propose that we introduce a new parameter, that we name `--triggers`, or something similar. This parameter would default to the value `schedule`, but would also allow the user to specify an array of allowed/enabled triggers.  Currently, the only two valid options would be `schedule` and `api`.

It would also allow us to drop the somewhat ambiguous `--enable-http-api` that currently coexist with `--enable-http-metrics` (although the only thing they have in common is that they both are served over HTTP).

This would leave us with enough flexibility to add more options going forward, allowing for instance triggering via message queues or similar.

@piksel @victorcmoura @zoispag @tammert, feel free to weigh in here.

piksel:
Actually, the `--run-once` could also be seen as `--triggers startup`? That would also as a side-effect remove `schedule` from the `triggers`, and cause it to exit automatically after the initial check since no more triggers could trigger.

DasSkelett:
> Just a thought, wouldn't manually specifying a schedule or interval warrant the enabling of the periodic updates as well?
> It would be breaking behaviour, but it seems like a more obvious way to interpret the command line/environment args.

Yup, that's how I originally planned to do it for #916, but when looking at the code I realized this is not possible with the current cmdline + environment parsing system, as @simskij already explained.

I like the ideas being expressed here, and as a user would find something like `--triggers` a very good improvement.
It would not only allow customizing the behaviour of Watchtower to your own needs (API triggers on/off, schedule on/off, startup on/off, and all possible combinations thereof), but also get rid of the unexpected side effects some commands currently have.
[...]

PR #1091, opened 2021-10-11
feat: improve HTTP API logging, honor no-startup-message

jinnatar:
They provide little useful information outside debugging and anything at
Info level is broadcast over shoutrrr.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -261,6 +261,7 @@ func formatDuration(d time.Duration) string {
 
 func writeStartupMessage(c *cobra.Command, sched time.Time, filtering string) {
 	noStartupMessage, _ := c.PersistentFlags().GetBool("no-startup-message")
+	enableUpdateAPI, _ := c.PersistentFlags().GetBool("http-api-update")
 
 	var startupLog *log.Entry
 	if noStartupMessage {
@@ -288,6 +289,11 @@ func writeStartupMessage(c *cobra.Command, sched time.Time, filtering string) {
 		startupLog.Info("Note that the first check will be performed in " + until)
 	} else {
 		startupLog.Info("Running a one time update.")
+	}
+
+	if enableUpdateAPI {
+		// TODO: make listen port configurable
+		startupLog.Info("The HTTP API is enabled at :8080.")
 	}
 
 	if !noStartupMessage {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -294,7 +294,7 @@ func writeStartupMessage(c *cobra.Command, sched time.Time, filtering string) {
 		startupLog.Info("Scheduling first run: " + sched.Format("2006-01-02 15:04:05 -0700 MST"))
 		startupLog.Info("Note that the first check will be performed in " + until)
 	} else {
-		startupLog.Info("Running a one time update.")
+		startupLog.Info("Periodic runs are not enabled.")
 	}
 
 	if enableUpdateAPI {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -159,6 +159,7 @@ func Run(c *cobra.Command, names []string) {
 	enableMetricsAPI, _ := c.PersistentFlags().GetBool("http-api-metrics")
 	unblockHTTPAPI, _ := c.PersistentFlags().GetBool("http-api-periodic-polls")
 	apiToken, _ := c.PersistentFlags().GetString("http-api-token")
+	enableAPIPeriodicPolls, _ := c.PersistentFlags().GetBool("http-api-periodic-polls")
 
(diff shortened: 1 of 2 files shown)


piksel:
I don't think it is that straight forward. It essentially replaces the startup message and when `--debug` isn't passed, there is no output from starting watchtower at all which might be confusing for some users. That being said, the startup message should probably be displayed, but incorporating whether the API is enabled. That way you could disable it as well using `--no-startup-message`.

simonaronsson (changes requested):
I agree with @piksel on this.

jinnatar:
ACK, reworking it to use `--no-startup-message`

jinnatar:
@piksel PTAL. I left the `runHTTPServer` logging as Debug since it's of questionable value to a user, and changed the `Info` level log to contain the listen port so it's more useful in indicating the API is functioning and where to find it.

Example run:
```
%> ./watchtower daemon --http-api-update --http-api-token foo --http-api-periodic-polls --no-startup-message
INFO[0001] Watchtower v0.0.0-unknown                     notify=no
INFO[0001] Using no notifications                        notify=no
INFO[0001] Only checking containers with name "daemon"   notify=no
INFO[0001] Scheduling first run: 2021-10-24 16:29:02 +0300 EEST  notify=no
INFO[0001] Note that the first check will be performed in 23 hours, 59 minutes, 59 seconds  notify=no
INFO[0001] The HTTP API is enabled at :8080.             notify=no
```

jinnatar:
There is still a slight annoyance left where watchtower is very silent and unhelpful, but I believe this is a deeper issue than I can resolve here:
```
%> ./watchtower daemon --http-api-update --http-api-token foo 
<absolutely no output produced>
```

i.e. if the API mode is enabled but no periodic updates are enabled as is the default, startup messages aren't processed at all since the call is blocked. Probably they should still happen, but then the flow needs to be changed a bit and my Go isn't exactly that good to do it in this PR without undue risk to general stability. :-)

Edit: nvm, found a clean way to fix this issue at the same time.

cmd/root.go
```diff
@@ -159,6 +159,7 @@ func Run(c *cobra.Command, names []string) {
 	enableMetricsAPI, _ := c.PersistentFlags().GetBool("http-api-metrics")
 	unblockHTTPAPI, _ := c.PersistentFlags().GetBool("http-api-periodic-polls")
 	apiToken, _ := c.PersistentFlags().GetString("http-api-token")
+	enableAPIPeriodicPolls, _ := c.PersistentFlags().GetBool("http-api-periodic-polls")
```
piksel: Why make another copy of the flag?

pkg/api/api.go
```diff
@@ -71,6 +70,6 @@ func (api *API) Start(block bool) error {
 }
 
 func runHTTPServer() {
-	log.Info("Serving HTTP")
+	log.Debug("Serving HTTP")
```
piksel:
```suggestion
```
I don't think this serves any purpose now

cmd/root.go
```diff
@@ -287,7 +294,12 @@ func writeStartupMessage(c *cobra.Command, sched time.Time, filtering string) {
 		startupLog.Info("Scheduling first run: " + sched.Format("2006-01-02 15:04:05 -0700 MST"))
 		startupLog.Info("Note that the first check will be performed in " + until)
 	} else {
-		startupLog.Info("Running a one time update.")
+		startupLog.Info("Periodic runs are not enabled.")
+	}
```
piksel, line 299: This is a bit weird when just starting with `--run-once`, as it will be followed by said update.

```suggestion
		if runOnce, _ := c.PersistentFlags().GetBool("run-once"); runOnce {
			startupLog.Info("Running a one time update.")
        	} else {
			startupLog.Info("Periodic runs are not enabled.")
		}
	}
```

It should probably be `else if` etc. but that's not supported by GitHub suggestions...

piksel (changes requested):
Looks good, except for a few smaller things.

The API parts are messy and should be redone (ref #939). But this should be sufficient for now. Just take a look at the comments.

cmd/root.go
jinnatar: Because I was being blind and didn't notice `unblockHTTPAPI ` was already there. Will fix.

cmd/root.go
jinnatar: It made sense to me, but you could be right that not to anyone else so better to be more helpful. Added the else if version.

piksel (approved):

jinnatar:
> I agree with @piksel on this.

I believe these have been addressed but GitHub still marks this comment as blocking merge for some reason.

simonaronsson (approved):

PR #1096, opened 2021-10-18
Adding a user option for the lifecycle hooks

patricegautier:
Hi all,

This pull request adds a user option for the lifecycle hooks, which enables them to be run under a particular user when needed, for example:

com.centurylinklabs.watchtower.lifecycle.pre-check.user="myuser"

that value is then passed to the docker command invocation for the given hook.

Thanks!


patricegautier:
forgot to update the test code - pulling this for now.

PR #1098, opened 2021-10-18
Add a user option for lifecycle hooks

patricegautier:
Hi all,

This pull request adds a user option for the lifecycle hooks, which enables them to be run under a particular user for example:

com.centurylinklabs.watchtower.lifecycle.pre-check.user="myuser"

that value is then passed to the docker exec command invocation for the given hook.

Thanks!

### DIFF
--- a/docs/lifecycle-hooks.md
+++ b/docs/lifecycle-hooks.md
@@ -50,6 +50,12 @@ the `docker run` command line:
     someimage --label=com.centurylinklabs.watchtower.lifecycle.post-check="/send-heartbeat.sh" \
     ```
 
+### Specifying a user for lifecycle hooks
+
+The lifecycle hooks all support a '.user' extension to specify a user to run the command with, for example:""
+
+com.centurylinklabs.watchtower.lifecycle.pre-check.user="myuser"
+
 ### Timeouts
 The timeout for all lifecycle commands is 60 seconds. After that, a timeout will
 occur, forcing Watchtower to continue the update loop.
--- a/internal/actions/mocks/client.go
+++ b/internal/actions/mocks/client.go
@@ -75,7 +75,7 @@ func (client MockClient) GetContainer(_ t.ContainerID) (container.Container, err
 }
 
 // ExecuteCommand is a mock method
-func (client MockClient) ExecuteCommand(_ t.ContainerID, command string, _ int) (SkipUpdate bool, err error) {
+func (client MockClient) ExecuteCommand(_ t.ContainerID, command string, user string, _ int) (SkipUpdate bool, err error) {
 	switch command {
 	case "/PreUpdateReturn0.sh":
 		return false, nil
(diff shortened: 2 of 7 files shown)


patricegautier:
Hi there - any chance of getting this pull request reviewed?  Thanks in advance

piksel:
You are still missing tests: ` 0.00% of diff hit (target 61.96%)`

patricegautier:
Hi @piksel - I added a teensy bit of test coverage in #def6f66 but to be honest I don’t quite know what I am doing there.. Would appreciate pointers on how to resolve this. Thanks in advance..

simonaronsson:
Would this be preferable to doing `su - someuser`? I  get the convenience, but at the same time it adds quite a bit of additional configuration parameters, which might not necessarily be good. @piksel, what is your opinion?

patricegautier:
In some cases you can’t do this from inside the container,  which is what this solves..

	-P

piksel:
Also, is running `docker exec --user foo container sh` the same as `docker exec container su foo`? Wouldn't the latter require a mounted `/etc/passwd`?

**Update:** They are, and require a user in `/etc/passwd`.
The only difference is that `--user` can be done with a `UID` instead.

```sh
❯ docker exec -it --user 1000 t1 whoami
whoami: unknown uid 1000

❯ docker exec -it t1 su 1000 -c whoami
su: unknown user 1000
```

@patricegautier, what do you mean by "can't"? That `su` (or often busybox) is missing in the container? Or that the image is built with `USER notroot`?

patricegautier:
> Also, is running `docker exec --user foo container sh` the same as `docker exec container su foo`? Wouldn't the latter require a mounted `/etc/passwd`?
> 
> **Update:** They are, and require a user in `/etc/passwd`. The only difference is that `--user` can be done with a `UID` instead.
> 
> ```shell
> ❯ docker exec -it --user 1000 t1 whoami
> whoami: unknown uid 1000
> 
> ❯ docker exec -it t1 su 1000 -c whoami
> su: unknown user 1000
> ```
> 
> @patricegautier, what do you mean by "can't"? That `su` (or often busybox) is missing in the container? Or that the image is built with `USER notroot`?

In the case I was trying to solve, the latter but I think I have seen the former as well..

patricegautier:
Also, the test runs failed with:

```
Ran 4 of 4 Specs in 0.003 seconds
312
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 0 Skipped
313
--- PASS: TestManifest (0.00s)
314
PASS
315
coverage: 80.0% of statements
316
ok  	github.com/containrrr/watchtower/pkg/registry/manifest	0.200s	coverage: 80.0% of statements
317
?   	github.com/containrrr/watchtower/pkg/session	[no test files]
318
?   	github.com/containrrr/watchtower/pkg/sorter	[no test files]
319
?   	github.com/containrrr/watchtower/pkg/types	[no test files]
```

which is not an area I think I touched.. Would appreciate pointers on what to do there.  Thanks

PR #1102, opened 2021-10-22
fixing flags usage text to first capital letter.

dhiemaz:
I see there is usage text which uses the first lower capital letter and the others are the first capital letter. This PR is Changing those flags usage text to first capital letter.


simonaronsson (approved):
Thank you for your contributions!

PR #1114, opened 2021-10-31
User life cycle hooks

patricegautier:
This is a continuation of 'Add a user option for lifecycle hooks #1098', which automatically got closed as I was doing some branch maintenance on my fork, sorry for the noise..


patricegautier:
as a little bonus the test that was failing is fixed.

patricegautier:
.. and I also added a post-update timeout option, symmetrical to the pre-update timeout one..

piksel:
I refactored the container client tests to make it possible for better testing of this (using ghttp).

> .. and I also added a post-update timeout option, symmetrical to the pre-update timeout one..

This might be a good PR on it's own, but it should probably be separated. What was the motivation for adding it here?

patricegautier:
> I refactored the container client tests to make it possible for better testing of this (using ghttp).

Thanks I will take a look.

> .. and I also added a post-update timeout option, symmetrical to the pre-update timeout one..
> 
> This might be a good PR on it's own, but it should probably be separated. What was the motivation for adding it here?

You are right it should have been separate.  Let me know if that's a must now that it's in there..

piksel:
> Let me know if that's a must now that it's in there..

That part will need some discussion as well and having it in the same PR thread will get messy. And also, every PR gets turned into a change list item, and "User life cycle hooks" gives no indication about timeouts (and should be it's own item).

patricegautier:
Ok no worries, I’ll take it out and create a separate PR.

	-P

patricegautier:
FYI - removed the post update timeout bit - I will submit a separate PR

patricegautier:
and post update timeout changes are now in #1124.  Thanks.

docs/lifecycle-hooks.md
```diff
@@ -50,13 +50,19 @@ the `docker run` command line:
...
+com.centurylinklabs.watchtower.lifecycle.pre-check.user="myuser"
+
 ### Timeouts
 The timeout for all lifecycle commands is 60 seconds. After that, a timeout will
 occur, forcing Watchtower to continue the update loop.
 
 #### Pre-update timeouts
 
-For the `pre-update` lifecycle command, it is possible to override this timeout to
+For the `pre-update` the lifecycle command, it is possible to override this timeout to
```
piksel:
```suggestion
For the `pre-update` lifecycle command, it is possible to override this timeout to
```

piksel (approved):
The docs revert accidently a word, but otherwise this looks good imo.

pkg/container/container_test.go
```diff
@@ -213,6 +213,28 @@ var _ = Describe("the container", func() {
 					Expect(links).To(SatisfyAll(ContainElement("redis"), ContainElement("postgres"), HaveLen(2)))
 				})
 			})
+			When("there is a pre or post update timeout", func() {
+				It("should return minute values", func() {
```
simonaronsson, line 217:
```
				It("should return minute values", func() {
```

copy/paste artifact?

simonaronsson (changes requested):

pkg/container/container_test.go
piksel: it applies to one of the env variables, but it should probably be split into 
```suggestion
				It("should return the minute value", func() {
```
and 
```go
			When("there is a pre or post update user", func() {
				It("should return the specified user", func() {
```

PR #1115, opened 2021-10-31
feat: use a more specific error type for no container info

MorrisLaw:
This PR adds a new error type to when `ContainerInfo` comes back nil while verifying the configuration. It currently returns `errorInvalidConfig`, this change will return `errorNoContainerInfo` instead. Similar to how we return `errorNoImageInfo` when `ImageInfo` is nil.


piksel (approved):
Can't argue with being more specific in error messages. Thanks!

Not so sure I'd call it a refactor though... 😅

PR #1116, opened 2021-10-31
fix(api): return appropriate status for unauthorized requests

### LINKED ISSUE #1095: Watchtower API Mode always returns 200 OK status code
**Describe the bug**
When using watchtower in API mode sending any request to `v1/update` endpoint it will always return status `200 OK`.

**To Reproduce**
1. Setup watchtower in API mode
2. Make a request to `http://ip:port/v1/update` with missing/invalid `Authorization` header
3. Look at response code

**Expected behavior**
The API returns `401 Unauthorized` whenever the `Authorization` header is missing/invalid

**Screenshots**
image

**Environment**
Unrelated

<details>
 <summary><b> Logs from running watchtower with the <code>--debug</code> option </b></summary>

```
"Updates triggered by HTTP API request." only triggered when providing a valid token
```

</details>

**Additional context**
Would be nice to have meaningful status codes instead of just silently failing

hypnoglow:
Fixes #1095

Also added the test for both cases 😊

### DIFF
--- a/pkg/api/api_test.go
+++ b/pkg/api/api_test.go
@@ -1,0 +1,65 @@
+package api
+
+import (
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	token  = "123123123"
+)
+
+func TestAPI(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "API Suite")
+}
+
+var _ = Describe("API", func() {
+	api := New(token)
+
+	Describe("RequireToken middleware", func() {
+		It("should return 401 Unauthorized when token is not provided", func() {
+			handlerFunc := api.RequireToken(testHandler)
+
+			rec := httptest.NewRecorder()
+			req := httptest.NewRequest("GET", "/hello", nil)
+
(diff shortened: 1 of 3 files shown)


simonaronsson:
Excellent work! Thank you! I'll have a more thorough look at the code as soon as I get an opportunity.

pkg/api/api_test.go
```diff
@@ -0,0 +1,65 @@
...
+			rec := httptest.NewRecorder()
+			req := httptest.NewRequest("GET", "/hello", nil)
+			req.Header.Set("Authorization", "Bearer 123")
+
+			handlerFunc(rec, req)
+
+			Expect(rec.Code).To(Equal(http.StatusUnauthorized))
+		})
+
+		It("should return 401 Unauthorized when token is invalid", func() {
```
piksel:
```suggestion
		It("should return 200 OK when token is valid", func() {
```

piksel (approved):
Looks really good. I have no idea how the commit manages to "decreases coverage by 0.46%" while also having 100% diff coverage...

pkg/api/api_test.go
hypnoglow: Nice catch, thanks. Fixed.

piksel (approved):

piksel:
@all-contributors please add @hypnoglow for code

PR #1146, opened 2021-12-01
docs: fix redirect link

jiangtj:
Fix redirect link: from https://containrrr.github.io to https://containrrr.dev


piksel (commented):
Good catch!

piksel (approved):

PR #1151, opened 2021-12-04
chore: bump alpine version in dockerfile

naioja:
Using a newer supported alpine version that still provides security fixes.


simonaronsson:
Not sure whether this is relevant given that we only use alpine to fetch the certs and the timezone. Wouldn't hurt though, so I guess why not 😅

naioja:
> Not sure whether this is relevant given that we only use alpine to fetch the certs and the timezone. Wouldn't hurt though, so I guess why not 😅

Running trivy on the old image was a bit unsettling :)

piksel (approved):

PR #1171, opened 2021-12-20
fix: move invalid token to log field

piksel:
Prevents potential log manipulation. Probably not a realistic concern, but fixing it is easier than dismissing the warnings.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -29,10 +29,6 @@ func (api *API) RequireToken(fn http.HandlerFunc) http.HandlerFunc {
 		auth := r.Header.Get("Authorization")
 		want := fmt.Sprintf("Bearer %s", api.Token)
 		if auth != want {
-			log.WithFields(log.Fields{
-				"got":      fmt.Sprintf("%q", auth),
-				"expected": want,
-			}).Tracef("Invalid Authorization header")
 			w.WriteHeader(http.StatusUnauthorized)
 			return
 		}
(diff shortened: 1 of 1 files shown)


piksel:
😩 

Welp, this PR actually does escape the input as a "safe" go-string, which should not trigger CWE-117 (all non-text characters are escaped, including newlines etc.).
So even if this fixes the potential issue, it still does not clear the warning (and even introduces another warning since we are outputting the HTTP headers).

simonaronsson:
Soo, are we closing this, or are you still looking to continue at some point? The easIER solution here would probably be to just escape it and log it as-is.

piksel:
Well, it *is* escaped (using %q). Not sure how we could escape it in a way that doesn't trigger the security rule. I think it's marginally better in this version, since it's clearly distinguished from an actual log message, but we could also just remove it all together.

simonaronsson:
I think I prefer the second option

pkg/api/api.go
```diff
@@ -28,8 +29,10 @@ func (api *API) RequireToken(fn http.HandlerFunc) http.HandlerFunc {
 		auth := r.Header.Get("Authorization")
 		want := fmt.Sprintf("Bearer %s", api.Token)
 		if auth != want {
-			log.Tracef("Invalid Authorization header \"%s\"", auth)
-			log.Tracef("Expected Authorization header to be \"%s\"", want)
+			log.WithFields(log.Fields{
+				"got":      fmt.Sprintf("%q", auth),
+				"expected": want,
+			}).Tracef("Invalid Authorization header")
```
simonaronsson:
```suggestion
```

simonaronsson (changes requested):

PR #1193, opened 2022-01-19
Add additional information for metrics.md

jamesmacwhite:
It might be me, but I thought that the metrics.md documentation could do with directly mentioning the specific endpoint that the Prometheus metrics are provided at and also providing an example scrape_config for Prometheus usage, if not using the demo example.

I had to end up going through the GitHub repository to find this information, so I thought it might be a good idea to include in the docs directly.


simonaronsson:
Thank you for your contribution! 🎉 
@all-contributors add @jamesmacwhite for docs

Issue #1127, opened 2021-11-11
Unknown flag --http-api-periodic-polls


pennal:
**Describe the bug**
According to the documentation at [https://containrrr.dev/watchtower/http-api-mode/](https://containrrr.dev/watchtower/http-api-mode/), when using the HTTP API mode the periodic checking is disabled. To re-enable it, one should use the `--http-api-periodic-polls` flag when running watchtower. 

Using the latest image I get the following error:
```
watchtower_1  | time="2021-11-11T14:20:58+01:00" level=fatal msg="unknown flag: --http-api-periodic-polls"
```

**To Reproduce**
Run the following docker-compose:

```yaml
version: "3"
services:
  watchtower:
    image: containrrr/watchtower 
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
    command: --http-api-periodic-polls
    environment:
      TZ: "Europe/Zurich"
      WATCHTOWER_DEBUG: "true"
      WATCHTOWER_HTTP_API_UPDATE: "true"
      WATCHTOWER_HTTP_API_TOKEN: "mytoken"
      WATCHTOWER_SCHEDULE: "0 */10 * ? * *"
    ports:
      - 8080:8080
```

**Expected behavior**
The HTTP API should be enabled, and periodic checks should also be enabled

**Screenshots**

**Environment**

Docker version 20.10.8, build 3967b7d
docker-compose version 1.29.2, build 5becea4c

<details>
 <summary><b> Logs from running watchtower with the <code>--debug</code> option </b></summary>

```
docker-compose up
Pulling watchtower (containrrr/watchtower:)...
latest: Pulling from containrrr/watchtower
e0c7a3e4c313: Pull complete
db1b75c6d82b: Pull complete
5bdc0dfc1598: Pull complete
Digest: sha256:3283e0b5be326d77ff4f4e8b7a91d46aaa1d511c74877b5a32f161548812d00c
Status: Downloaded newer image for containrrr/watchtower:latest
Creating desktop_watchtower_1 ... done
Attaching to desktop_watchtower_1
watchtower_1  | Error: unknown flag: --http-api-periodic-polls
watchtower_1  | Usage:
watchtower_1  |   watchtower [flags]
watchtower_1  |
watchtower_1  | Flags:
watchtower_1  |   -a, --api-version string                          api version to use by docker client (default "1.25")
watchtower_1  |   -c, --cleanup                                     remove previously used images after updating
[...]

PR #1195, opened 2022-01-21
Fix typo on --http-api-update environment variable and add warning note for --http-api-periodic-polls

jamesmacwhite:
Minor docs issue, but the --http-api-update parameter has a typo on it's env var equivalent. Using the env var provided by the docs will not enable the HTTP API update only endpoint and you'll get a 404.

~~In addition due to #1127, the docs are currently listing a parameter which isn't actually available in any official release currently, so I thought a warning may be useful until a new release is made, which can then be removed.~~

Thanks!

### DIFF
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -241,7 +241,7 @@ For details see [HTTP API](https://containrrr.dev/watchtower/http-api-mode).
 
 ```text
             Argument: --http-api-update
-Environment Variable: WATCHTOWER_HTTP_API
+Environment Variable: WATCHTOWER_HTTP_API_UPDATE
                 Type: Boolean
              Default: false
 ```
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -257,6 +257,10 @@ Environment Variable: WATCHTOWER_HTTP_API_TOKEN
 ```
 
 ## HTTP API periodic polls
+
+!!! warning
+This argument is not currently available in the 1.3.0 release.
+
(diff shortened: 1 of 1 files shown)


docs/arguments.md
```diff
@@ -257,6 +257,10 @@ Environment Variable: WATCHTOWER_HTTP_API_TOKEN
 ```
 
 ## HTTP API periodic polls
+
+!!! warning
+This argument is not currently available in the 1.3.0 release.
+
```
piksel: The plan is to release today, so this won't be deployed until the update anyway.

```suggestion
```

docs/arguments.md
```diff
@@ -241,7 +241,7 @@ For details see [HTTP API](https://containrrr.dev/watchtower/http-api-mode).
 
 ```text
             Argument: --http-api-update
-Environment Variable: WATCHTOWER_HTTP_API
+Environment Variable: WATCHTOWER_HTTP_API_UPDATE
```
piksel, line 244: Correct. There is no `WATCHTOWER_HTTP_API` environment variable referenced in `flags.go`.

piksel (approved):

docs/arguments.md
jamesmacwhite: Ah OK, would you like me to remove this commit?

docs/arguments.md
jamesmacwhite: Thanks for confirming, yes it looked like a typo to me.

docs/arguments.md
simonaronsson: No worries, I'll take care of it as we're releasing today :)

PR #1208, opened 2022-01-28
Initial work to add Update Delay feature

### LINKED ISSUE #426: Delayed updates

It happens that an image is released, then recalled / pulled back because of an issue. Currently such an image would be updated by watchtower (as designed) 

The ability to delay the update, such as in *update the image, but only if it is 3 days old*

Ideally on a per image basis, but a global setting is ok as well. 

**Describe alternatives you've considered**

There are none I can think of, except a very long time between checks but this will not solve the issue of a faulty image is released right before the check (it will be even worse, as a possible recovery/corrected one will not be pulled quickly)

cdloh:
Work not complete but my initial attempt at adding an update delay feature

Lets you not update an image if X amount of time has passed since the new image was created

Note that this is all based off the Created Time field from the image. So if an image is published with an incorrect date it could be considered old enough.

Further work to be done but I'm looking for initial feedback before I spend more time on it.

Closes #426 properly

### DIFF
--- a/internal/actions/mocks/client.go
+++ b/internal/actions/mocks/client.go
@@ -94,3 +94,8 @@ func (client MockClient) IsContainerStale(_ container.Container) (bool, t.ImageI
 func (client MockClient) WarnOnHeadPullFailed(_ container.Container) bool {
 	return true
 }
+
+// always return true in the mock
+func (client MockClient) OldEnoughImage(_ container.Container, _ t.ImageID) bool {
+	return true
+}
--- a/internal/actions/update.go
+++ b/internal/actions/update.go
@@ -34,8 +34,9 @@ func Update(client container.Client, params types.UpdateParams) (types.Report, e
 
 	for i, targetContainer := range containers {
 		stale, newestImage, err := client.IsContainerStale(targetContainer)
-		shouldUpdate := stale && !params.NoRestart && !params.MonitorOnly && !targetContainer.IsMonitorOnly()
-		if err == nil && shouldUpdate {
+		oldEnough := client.OldEnoughImage(targetContainer, newestImage)
+		shouldUpdate := stale && !params.NoRestart && !params.MonitorOnly && !targetContainer.IsMonitorOnly() && oldEnough
+		if err == nil && shouldUpdate && oldEnough {
 			// Check to make sure we have all the necessary information for recreating the container
 			err = targetContainer.VerifyConfiguration()
 			// If the image information is incomplete and trace logging is enabled, log it for further diagnosis
@@ -58,6 +59,7 @@ func Update(client container.Client, params types.UpdateParams) (types.Report, e
 			progress.AddScanned(targetContainer, newestImage)
 		}
 		containers[i].Stale = stale
+		containers[i].OldEnoughImage = oldEnough
 
 		if stale {
 			staleCount++
@@ -74,7 +76,7 @@ func Update(client container.Client, params types.UpdateParams) (types.Report, e
 	var containersToUpdate []container.Container
 	if !params.MonitorOnly {
 		for _, c := range containers {
-			if !c.IsMonitorOnly() {
+			if !c.IsMonitorOnly() && c.OldEnoughImage {
 				containersToUpdate = append(containersToUpdate, c)
 				progress.MarkForUpdate(c.ID())
 			}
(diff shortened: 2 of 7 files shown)


internal/actions/mocks/client.go
```diff
@@ -94,3 +94,8 @@ func (client MockClient) IsContainerStale(_ container.Container) (bool, t.ImageI
 func (client MockClient) WarnOnHeadPullFailed(_ container.Container) bool {
 	return true
 }
+
+// always return true in the mock
```
piksel, line 98:
```suggestion
// OldEnoughImage always returns true in the mock
```

internal/flags/flags.go
```diff
@@ -45,6 +45,12 @@ func RegisterSystemFlags(rootCmd *cobra.Command) {
 		viper.GetDuration("WATCHTOWER_TIMEOUT"),
 		"Timeout before a container is forcefully stopped")
 
+	flags.DurationP(
+		"update-delay",
+		"",
+		viper.GetDuration("WATCHTOWER_UPDATE_DELAY"),
+		"Delay an image pull until it is atleast this old")
```
piksel, line 52:
```suggestion
		"Delay an image pull until it is at least this old")
```

pkg/container/metadata.go
```diff
@@ -1,18 +1,19 @@
...
+	monitorOnlyLabel       = "com.centurylinklabs.watchtower.monitor-only"
+	dependsOnLabel         = "com.centurylinklabs.watchtower.depends-on"
+	updateDelayLabel       = "com.centurylinklabs.watchtower.update-delay"
+	zodiacLabel            = "com.centurylinklabs.zodiac.original-image"
+	scope                  = "com.centurylinklabs.watchtower.scope"
+	preCheckLabel          = "com.centurylinklabs.watchtower.lifecycle.pre-check"
+	postCheckLabel         = "com.centurylinklabs.watchtower.lifecycle.post-check"
+	preUpdateLabel         = "com.centurylinklabs.watchtower.lifecycle.pre-update"
+	postUpdateLabel        = "com.centurylinklabs.watchtower.lifecycle.post-update"
+	preUpdateTimeoutLabel  = "com.centurylinklabs.watchtower.lifecycle.pre-update-timeout"
```
piksel, line 16: Just a note that this is auto-formatted due to the next line which probably wasn't added by something running gopls/go fmt, and only contains the added `com.centurylinklabs.watchtower.update-delay` label.

pkg/container/client.go
```diff
@@ -292,6 +295,39 @@ func (client dockerClient) HasNewImage(ctx context.Context, container Container)
...
+	ctx := context.Background()
+
+	updateDelay := client.updateDelay
+	if containerDelay, ok := container.getLabelValue(updateDelayLabel); ok {
+		labelDelay, err := time.ParseDuration(containerDelay)
+
+		// Should we return false or true here?
+		if err != nil {
+			log.Debugf("Update Delay label is invalid for %s", container.Name())
+			return false
```
piksel, line 308:
```suggestion
		// Should we return false or true here?
		if err != nil {
			log.WithFields(log.Fields{
				"value", containerDelay,
				"error", err
			}).Warnf("Update delay label is invalid for %q: %s", container.Name(), err)
			return false
```
If someone explicitly tried to prevent the container from being updated too early, skipping it and including the warning in the notification/log is probably the safest bet (returning false).

piksel (changes requested):
The general implementation seems good. The comments starting with the function name is a go thing 😁.

Might need some more tests though...

cdloh:
Just letting you know I haven't stopped working on this. Just haven't had a chance to get back to it yet. Hoping to have some cycles next week.

internal/actions/update.go
```diff
@@ -34,8 +34,9 @@ func Update(client container.Client, params types.UpdateParams) (types.Report, e
 
 	for i, targetContainer := range containers {
 		stale, newestImage, err := client.IsContainerStale(targetContainer)
-		shouldUpdate := stale && !params.NoRestart && !params.MonitorOnly && !targetContainer.IsMonitorOnly()
-		if err == nil && shouldUpdate {
+		oldEnough := client.OldEnoughImage(targetContainer, newestImage)
+		shouldUpdate := stale && !params.NoRestart && !params.MonitorOnly && !targetContainer.IsMonitorOnly() && oldEnough
+		if err == nil && shouldUpdate && oldEnough {
```
seeReadCode, line 39: `&& oldEnough` is used in `if` despite being included in `shouldUpdate` – so unnecessary here

seeReadCode:
@cdloh are you still planning on getting into this?  I'd be curious to know what remains and might be able to put some time into it. Thanks!

pannal:
Hey, is anything like this planned still? I'd love such a feature.

PR #1219, opened 2022-02-11
fix(notifications): title customization

### LINKED ISSUE #1205: env. WATCHTOWER_NOTIFICATION_EMAIL_SUBJECTTAG isn't working anymore since 1.4.0
**Describe the bug**
Since the update to 1.4.0, the environment variable WATCHTOWER_NOTIFICATION_EMAIL_SUBJECTTAG isn't working.
In my docker-compose.yml file, I set this : `- WATCHTOWER_NOTIFICATION_EMAIL_SUBJECTTAG=[ Watchtower-DS920+ ]`
But all email I received has this subject:
`Watchtower updates on 5d11f1abc714`
image

Before the update, I got :
image

**To Reproduce**

**Expected behavior**
It should be with the subject I declare in the env variables, like this not too old email:
image

**Screenshots**
See previous...

**Environment**
 - Platform = Synology NAS DS920+
 - Architecture = amd64 (CPU = Intel Celeron J4125)
 - Docker version = 20.10.3, build b455053
 
I use Portainer CE 2.11 to create my container

<details>
 <summary><b> Logs from running watchtower with the <code>--debug</code> option </b></summary>
I'm not sure how to do the --debug option...
But here what I have:
```log
time="2022-01-25T09:56:09+01:00" level=debug
time="2022-01-25T09:56:09+01:00" level=debug msg="Sleeping for a second to ensure the docker api client has been properly initialized."
time="2022-01-25T09:56:10+01:00" level=debug msg="Making sure everything is sane before starting"
time="2022-01-25T09:56:10+01:00" level=debug msg="Retrieving running containers"
time="2022-01-25T09:56:10+01:00" level=debug msg="There are no additional watchtower containers"
time="2022-01-25T09:56:10+01:00" level=debug msg="Watchtower HTTP API skipped."
time="2022-01-25T09:56:10+01:00" level=info msg="Watchtower 1.4.0"
time="2022-01-25T09:56:10+01:00" level=info msg="Using notifications: smtp"
time="2022-01-25T09:56:10+01:00" level=info msg="Only checking containers using enable label"
time="2022-01-25T09:56:10+01:00" level=info msg="Scheduling first run: 2022-01-26 05:00:00 +0100 CET"
time="2022-01-25T09:56:10+01:00" level=info msg="Note that the first check will be performed in 19 hours, 3 minutes, 49 seconds"
```

</details>

**Additional context**

### LINKED ISSUE #1207: Customize or suppress title in notification
So the 1.4.0 release added the title line "Watchtower updates on `container hostname`" to each notification. Adding &title= no longer works for the discord webhook url. I tried changing the notification template as well to the default but it still prints it. Is there a way to customize or suppress this message like it was in 1.3.0?

_Originally posted by @ksurl in https://github.com/containrrr/watchtower/discussions/1197#discussioncomment-2052226_

piksel:
This PR introduces two new flags to watchtower:
`--notification-skip-title`
and
`--notification-title-tag <TAG>` (with a backwards compatible alias to `notification-email-subjecttag`).

The `title-tag` flag works the same way as the `email-subjecttag` used to work. If supplied, it will prefix the titles for all notifications with `[<TAG>] `. Since it is generic among all notifications (which support titles), it was renamed to reflect this, but using the old name still works for now. Fixes #1205 

Supplying the `skip-title` flag will not override the `title` param for the shoutrrr service, so it can still be set to whatever value you want using the `&title=` query parameter of the shoutrrr URL. This should in effect remove the title if  no such query param is supplied. Fixes #1207

This was done in a combined effort to simplify the title generation and fixing the unintended breaking changes in 1.4.0

### DIFF
--- a/pkg/notifications/shoutrrr.go
+++ b/pkg/notifications/shoutrrr.go
@@ -58,7 +58,7 @@ type shoutrrrTypeNotifier struct {
 	done           chan bool
 	legacyTemplate bool
 	params         *types.Params
-	hostname       string
+	data           StaticData
 }
 
 // GetScheme returns the scheme part of a Shoutrrr URL
@@ -79,11 +79,9 @@ func (n *shoutrrrTypeNotifier) GetNames() []string {
 	return names
 }
 
-func newShoutrrrNotifier(tplString string, acceptedLogLevels []log.Level, legacy bool, hostname string, delay time.Duration, urls ...string) t.Notifier {
+func newShoutrrrNotifier(tplString string, acceptedLogLevels []log.Level, legacy bool, data StaticData, delay time.Duration, urls ...string) t.Notifier {
 
-	notifier := createNotifier(urls, acceptedLogLevels, tplString, legacy)
-	notifier.hostname = hostname
-	notifier.params = &types.Params{"title": GetTitle(hostname)}
+	notifier := createNotifier(urls, acceptedLogLevels, tplString, legacy, data)
 	log.AddHook(notifier)
 
 	// Do the sending in a separate goroutine so we don't block the main process.
@@ -92,7 +90,7 @@ func newShoutrrrNotifier(tplString string, acceptedLogLevels []log.Level, legacy
 	return notifier
 }
 
-func createNotifier(urls []string, levels []log.Level, tplString string, legacy bool) *shoutrrrTypeNotifier {
+func createNotifier(urls []string, levels []log.Level, tplString string, legacy bool, data StaticData) *shoutrrrTypeNotifier {
 	tpl, err := getShoutrrrTemplate(tplString, legacy)
 	if err != nil {
 		log.Errorf("Could not use configured notification template: %s. Using default template", err)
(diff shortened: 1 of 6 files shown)


pkg/notifications/shoutrrr.go
```diff
@@ -240,10 +243,15 @@ func getShoutrrrTemplate(tplString string, legacy bool) (tpl *template.Template,
 	return
 }
 
+// StaticData is the part of the notification template data model set upon initialization
+type StaticData struct {
+	Title string
+	Host  string
+}
```
simonaronsson, line 250: Could this naming be improved to clarify what it is? `StaticData` is very generic.

simonaronsson (approved):
I'm 💯  fine with merging this as-is but left a minor comment.

pkg/notifications/shoutrrr.go
piksel: It is, but it's the best way I found to describe it. It's the non-changing (static) part of the data used for notifications. It could perhaps be called `StaticTemplateData`, but it's not, *strictly speaking*, only used for that.
Any suggestions?

PR #1241, opened 2022-03-06
feat: regex container name filtering

mateuszdrab:
This small change allows the use of re2 regex in the container name filter parameter

I'm currently running this in my environment with the following cmdline
```--cleanup --interval 21600 -d \/[^k][^8][^s][^_][\S]+```

This prevents containers belonging to K8s from being auto updated as there is no way to enforce labels correctly in K8s.

I've been running this since end of January and had no problems so far - thought it would be nice to PR this.

### DIFF
--- a/pkg/filters/filters.go
+++ b/pkg/filters/filters.go
@@ -1,9 +1,10 @@
 package filters
 
 import (
-	t "github.com/containrrr/watchtower/pkg/types"
-	"strings"
 	"regexp"
+	"strings"
+
+	t "github.com/containrrr/watchtower/pkg/types"
 )
 
 // WatchtowerContainersFilter filters only watchtower containers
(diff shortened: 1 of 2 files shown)


mateuszdrab:
Should we merge it or leave it?

simonaronsson:
Sorry Mateusz, I've been preoccupied with other things for a while. Will have a look as soon as time allows. Thank you for your contribution. 🙏🏼

piksel:
Soo, to make this change backwards-compatible, we need to ensure that the whole name matches the regex, otherwise we would have containers which included the name as a substring match.
I added some tests to confirm this, and replaced the `regexp.MatchString`s with `*Regex.FindStringIndex` and verifying the match indices. 
Since the initial `/` is optional, the match needs to start at 0 or 1.

piksel (approved):
I have handled the scenarios that I was worried about. @simskij ?

mateuszdrab:
@piksel 

Thanks for merging

Quick question, will there be a release any time soon?

If not, I'm happy to use the dev builds in the meantime; however, those are only built for amd64 - any chance to have cross platform dev builds?

Thanks

piksel:
@mateuszdrab What platform do you need? `arm64`? If it's not giving me too much trouble it would of course be better to just do a multi-platform image, but otherwise I could just add a platform specific tag.

I have been piling up work for some time, so I can't give you a release date unfortunately.

mateuszdrab:
> @mateuszdrab What platform do you need? `arm64`? If it's not giving me too much trouble it would of course be better to just do a multi-platform image, but otherwise I could just add a platform specific tag.
> 
> I have been piling up work for some time, so I can't give you a release date unfortunately.

@piksel 
armv7 actually, haven't switched the Raspberry Pis to 64 bit yet since its a mixture of 2s, 3s and a 4.

PR #1263, opened 2022-04-03
feat(api): supports custom port with 8080 as default

fynxiu:
This is all about the custom port for the "http api server"


simonaronsson:
Thank you for the PR!

One question though: Why don't you just do some other port mapping at deploy time?

fynxiu:
I happened to spend some time skimming over the source code of watchtower and happened to see that there was a to-do in root.go. So, whatever, thanks for providing this practical tool.

Issue #1285, opened 2022-05-09
Update containers from registry notification


Foxite:
**Is your feature request related to a problem? Please describe.**
My own apps are built by CI and pushed to a private registry (`registry` container). Right now I have Watchtower setup to check this repository every minute, which is obviously very inefficient, but I found out that `registry` can send HTTP notifications, and Watchtower has an HTTP API. Perfect, except `registry` sends a notification for every layer that gets pushed (which can be several for a single `docker push`), and Watchtower checks every single container upon an API request (when we already know which ones need to be checked).

**Describe the solution you'd like**
To effectively automate this, at least two things need to be implemented:
- Add an option to "debounce" API calls so that the first API call will trigger an update after a specified delay, and every other API call before the delay is over will be ignored.
- Add an API endpoint that allows you to specify which image should be checked (not container, but image, if possible).
- If possible, add an API endpoint that specificially gets this information from the notification sent by `registry`, so there doesn't need to be a script to translate a push notification to an API call.

**Describe alternatives you've considered**
An alternative to the first request is for someone fluent in Go (not me, unfortunately) to send a PR to https://github.com/distribution/distribution and have it debounce its notifications, which are sent in large volumes anyway.

An alternative to the second/third request is to have a separate, scoped Watchtower instance for every container you want to update in this way. Not practical if you have more than a few apps, and since `registry` appears to only support one notification endpoint, you'd need to have an "intermediate" API that selects the right Watchtower instance based on which image got pushed. (If you did this, you might as well have it debounce notifications.)

Foxite:
Currently working on this, but I have never written any Go so it'll take a while and probably be sub-par code, but hey it's a learning experience :upside_down_face:

piksel:
The simplest solution to this I can think of is to just add a "minimum interval" / cooldown argument that will skip updates until a set amount of time has passed.
It would still do a normal checking session though (ignoring the content of the push body), but it wouldn't update any containers that didn't have any new images of course.
We generally try to avoid adding vendor-specific handling (like parsing the update payload from distribution registry), but a generic API endpoint that accepts filters for a session (like container names in this example) should be fine.
You would still need some glue proxy that would translate the push payload to the watchtower API though.

Foxite:
Done by #1289

PR #1289, opened 2022-05-12
feat(http): optional query parameter to update only containers of a specified image

Foxite:
Implements the second request from #1285: Added an optional query parameter that allows you to only update containers of the specified image. This works even if the image was built locally (ie the container no longer has the image label in `docker ps`).

Also includes an indentation fix that Goland went on strike over.

I will implement the first request later today, and probably add the ability to specify more than one image as well. This draft PR is so there can be a code review of what's already been done.

Don't hold back on roasting my Go skills. :P

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -189,7 +189,7 @@ func Run(c *cobra.Command, names []string) {
 	httpAPI := api.New(apiToken)
 
 	if enableUpdateAPI {
-		updateHandler := update.New(func() { runUpdatesWithNotifications(filter) }, updateLock)
+		updateHandler := update.New(func(tags []string) { runUpdatesWithNotifications(filters.FilterByImageTag(tags, filter)) }, updateLock)
 		httpAPI.RegisterFunc(updateHandler.Path, updateHandler.Handle)
 		// If polling isn't enabled the scheduler is never started and
 		// we need to trigger the startup messages manually.
@@ -293,7 +293,7 @@ func writeStartupMessage(c *cobra.Command, sched time.Time, filtering string) {
 		startupLog.Info("Scheduling first run: " + sched.Format("2006-01-02 15:04:05 -0700 MST"))
 		startupLog.Info("Note that the first check will be performed in " + until)
 	} else if runOnce, _ := c.PersistentFlags().GetBool("run-once"); runOnce {
-			startupLog.Info("Running a one time update.")
+		startupLog.Info("Running a one time update.")
 	} else {
 		startupLog.Info("Periodic runs are not enabled.")
 	}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -189,7 +189,7 @@ func Run(c *cobra.Command, names []string) {
 	httpAPI := api.New(apiToken)
 
 	if enableUpdateAPI {
-		updateHandler := update.New(func(tags []string) { runUpdatesWithNotifications(filters.FilterByImageTag(tags, filter)) }, updateLock)
+		updateHandler := update.New(func(images []string) { runUpdatesWithNotifications(filters.FilterByImage(images, filter)) }, updateLock)
 		httpAPI.RegisterFunc(updateHandler.Path, updateHandler.Handle)
 		// If polling isn't enabled the scheduler is never started and
 		// we need to trigger the startup messages manually.
--- a/pkg/api/update/update.go
+++ b/pkg/api/update/update.go
@@ -13,7 +13,7 @@ var (
 )
 
 // New is a factory function creating a new  Handler instance
-func New(updateFn func(), updateLock chan bool) *Handler {
+func New(updateFn func(images []string), updateLock chan bool) *Handler {
 	if updateLock != nil {
 		lock = updateLock
 	} else {
@@ -29,7 +29,7 @@ func New(updateFn func(), updateLock chan bool) *Handler {
 
 // Handler is an API handler used for triggering container update scans
 type Handler struct {
-	fn   func()
+	fn   func(images []string)
 	Path string
 }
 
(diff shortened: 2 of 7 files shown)


Foxite:
CI is saying url.Has is undefined, but this very much compiles (and works) on my end. Am I using the wrong Go version or something? I didn't tell my IDE what it should use, go.mod seems to set it to 1.12 which doesn't match what the CI seems to use (1.15).

Foxite:
In any case, I think it's probably better to implement debouncing in the "glue proxy" that's needed if you use this to trigger updates by registry notifications, which is the only reason why debouncing is necessary in the first place.

pkg/api/update/update.go
```diff
@@ -43,12 +44,15 @@ func (handle *Handler) Handle(w http.ResponseWriter, r *http.Request) {
...
 
-	select {
-	case chanValue := <-lock:
-		defer func() { lock <- chanValue }()
-		handle.fn()
-	default:
-		log.Debug("Skipped. Another update already running.")
+	var images []string
+	if r.URL.Query().Has("image") {
+		images = strings.Split(r.URL.Query().Get("image"), ",")
```
piksel: Since the URL query technically allows setting the same key multiple times, this would also allow for both the `?image=a&image=b` as well as the `?image=a,b` style:

```suggestion
	if imageQueries, found := r.URL.Query()["image"] {
		for _, image := range imageQueries {
			images = append(images, strings.Split(image, ","))
		}
```

If you only want to support the comma style, you could instead simply do:
```suggestion
	if image := r.URL.Query().Get("image"); image != "" {
		images = strings.Split(image, ",")
```

piksel:
> CI is saying url.Has is undefined, but this very much compiles (and works) on my end. Am I using the wrong Go version or something? I didn't tell my IDE what it should use, go.mod seems to set it to 1.12 which doesn't match what the CI seems to use (1.15).

Yeah, 
> func (Values) [Has](https://pkg.go.dev/net/url@go1.18.2#Values.Has), added in go1.17

But I added a suggestion for skipping it (and saving a few CPU cycles! :D)

pkg/api/update/update.go
```diff
@@ -43,12 +44,15 @@ func (handle *Handler) Handle(w http.ResponseWriter, r *http.Request) {
...
+	} else {
+		images = nil
 	}
 
+	chanValue := <-lock
+	defer func() { lock <- chanValue }()
+	handle.fn(images)
```
piksel: This still need to check if the lock is taken before running the update, otherwise it's kind of pointless to have a lock...
```suggestion
select {
case chanValue := <-lock:
	defer func() { lock <- chanValue }()
	handle.fn(images)
default:
	log.Debug("Skipped. Another update already running.")
}
```

cmd/root.go
```diff
@@ -189,7 +189,7 @@ func Run(c *cobra.Command, names []string) {
 	httpAPI := api.New(apiToken)
 
 	if enableUpdateAPI {
-		updateHandler := update.New(func() { runUpdatesWithNotifications(filter) }, updateLock)
+		updateHandler := update.New(func(tags []string) { runUpdatesWithNotifications(filters.FilterByImageTag(tags, filter)) }, updateLock)
```
piksel:
```suggestion
		updateHandler := update.New(func(images []string) { runUpdatesWithNotifications(filters.FilterByImageTag(images, filter)) }, updateLock)
```
We should be consistent, but perhaps even `imageTags` is the most opaque term to use...

piksel (changes requested):
In general, this looks good! The main blocker is the lock handling (and the URL.Values.Has call).

pkg/api/update/update.go
Foxite: I could be wrong but I understood the original code to skip an update altogether when one is already running, while my change would make it wait until it's done before proceeding with the new update. Although I didn't actually look this up.

My idea was that you can issue multiple update requests in short succession, for different images (as a busy CI system might do), and they won't interfere with each other.

cmd/root.go
Foxite: So I understand that you think it should be changed to `imageTags` everywhere, is that correct?

Foxite: I've spotted the difference between "image" and "tag"; I've changed the parameter/variable names to image.

pkg/api/update/update.go
Foxite: There was a compile error in here, I'm still not sure if this is because my Go version is incorrect, even though I've now explicitly told my IDE to use 1.15.

pkg/api/update/update.go
Foxite: @piksel so my change makes the lock useless because it proceeds regardless of whether or not the lock was taken, but we should still allow multiple update requests for different images to be queued, no? I don't know how to do that, though, but I could find out of course.

piksel: If the channel is sized to 1, it will block until the lock can be acquired. It does make more sense to queue up the requests now when they might have different parameters. I think it's generally fine, but since it's changed behaviour, maybe we can check if `len(images) > 0` and only block if that is the case. This is just to prevent any previously working setups from suddenly start checking continuously.

piksel: If the channel is sized to 1, it will block until the lock can be acquired. It does make more sense to queue up the requests now when they might have different parameters. I think it's generally fine, but since it's changed behaviour, maybe we can check if `len(images) > 0` and only block if that is the case. This is just to prevent any previously working setups from suddenly start checking continuously.

piksel: (I get that this was what your original code was doing, I just misunderstood what you meant with "useless" above. In my suggestion it should've been followed by a `return`)

piksel:
@Foxite are you still working on this?

Foxite:
I was confused by your earlier comment about the locking, I thought it was going to be complicated to make it do what I want (and postponed the task until I have time to learn about it). But upon re-reading it it seems simple; just to be clear, I understand that all that needs to be done is check if any images are specified, and only block on the lock if that's the case -- if not, then skip the request if the lock is not immediately acquired.

Or:

```go
if len(images) > 0 {
  // old locking code
} else {
  // new locking code
}
```
Is that correct?

Foxite:
I've gone ahead and done that, and it works. So if I've understood you correctly, this should be ready to merge.

piksel:
Sorry about that. Yeah, that was what I meant. When there are filter arguments to the requests, it makes sense to queue them up (unless they are the same, but would be much more complex to implement), but if it's just a "do the update now" it's probably better to ignore them ("yes, I am currently doing the update"). And also for the aforementioned backwards compatibility.

Foxite:
Looks like code coverage target is not met, I should probably write unit tests.

Foxite:
@piksel I have added unit tests and all the checks are green, how does it look?

piksel (approved):
LGTM! Great job.

piksel:
@all-contributors add @Foxite for code

PR #1300, opened 2022-06-06
Support secrets for notification_url

jlaska:
Since the NOTIFICATION_URL contains app and user defined tokens, this PR adds support for the parameter to be loaded from a secret file.  Relevant documentation updated to note that a file path is supported.

### DIFF
--- a/docs/notifications.md
+++ b/docs/notifications.md
@@ -181,7 +181,8 @@ If you want to disable TLS verification for the Gotify instance, you can use eit
 
 To send notifications via shoutrrr, the following command-line options, or their corresponding environment variables, can be set:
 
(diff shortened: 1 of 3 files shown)


piksel:
This used to not be possible to do due to a bug in cobra/viper/pflags. It should have been fixed now, so this might work as expected. Have you tried it with multiple values etc.?

piksel:
No, the problem is still there. The `StringArray` flags only appends the values passed to `Set`, which means that additionally to the file contents, there will always be the file name (which probably isn't a valid notification URL).
There is a way to update this correctly though, I will ~~make a new PR for this~~ amend the PR.

piksel (approved):

jlaska:
This is great, thank you @piksel !

PR #1303, opened 2022-06-09
fix(notifications): include icon in slack legacy url

Choromanski:
#### Modified parsing of discord URLs:
 * The previous implementation assumed that the discord URL had a trailing path argument in the URL path
   * Ex: `https://discordapp.com/api/webhooks/WEBHOOK_ID/TOKEN/slack`
* However discord generates webhook URLs without the trailing path argument
  * Ex: `https://discordapp.com/api/webhooks/WEBHOOK_ID/TOKEN`
* Seeing as the `/slack` argument is never parsed by watchtower or sent to shoutrrr it should not be a requirement
* This PR addresses this issue by supporting URLs of both types:
  * `https://discordapp.com/api/webhooks/WEBHOOK_ID/TOKEN/slack` and `https://discordapp.com/api/webhooks/WEBHOOK_ID/TOKEN`

#### Added support for the `--notification-slack-icon-url` argument when using a discord webhook:
 * Previously this argument did nothing when using a discord webhook URL
 * Now it is parsed as the Avatar in the shoutrrr discord config

#### Added two tests for my additions
 * One to test the parsing of the `WEBHOOK_ID` and `TOKEN` when there is no trailing path argument in the discord webhook URL
 * One to test the parsing of the `--notification-slack-icon-url` argument with a discord webhook URL
   * This also tests the parsing of the `--notification-slack-identifier` argument with a discord webhook URL (not previously tested)

### DIFF
--- a/pkg/notifications/notifier_test.go
+++ b/pkg/notifications/notifier_test.go
@@ -149,6 +149,8 @@ var _ = Describe("notifications", func() {
 			color := notifications.ColorInt
 			data := notifications.GetTemplateData(command)
 			title := url.QueryEscape(data.Title)
+			username := "containrrrbot"
+			iconURL := "https://containrrr.dev/watchtower-sq180.png"
 			expected := fmt.Sprintf("discord://%s@%s?color=0x%x&colordebug=0x0&colorerror=0x0&colorinfo=0x0&colorwarn=0x0&title=%s&username=watchtower", token, channel, color, title)
 			buildArgs := func(url string) []string {
 				return []string{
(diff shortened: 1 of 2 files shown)


piksel (commented):
> The previous implementation assumed that the discord URL had a trailing path argument in the URL path
Ex: https://discordapp.com/api/webhooks/WEBHOOK_ID/TOKEN/slack

Yes, this is because that is the slack API emulation endpoint for discord. Using the regular discord webhook API with a slack client would not work.
Proper discord support is added through shoutrrr, and this is only a helper that creates a shoutrrr URL from the arguments. It should behave as the original implementation.

> Added support for the --notification-slack-icon-url 

This was actually an unintended regression, so this should probably be added.

Choromanski:
Looking through the parsing of the hookURL in the GetURL function the /slack endpoint is never parsed and therefore makes no difference when it is present or not. I however have not looked through the shoutrrr source and would not be surprised if shoutrrr is adding the /slack endpoint back in.

I made this change to help simplify discord notification integration. IMHO when you are given a webhook URL you shouldn't have to modify the URL to use it.

The lack of documentation when it comes to using discord notifications requires digging into the source to determine that the webhook URL needs an extra endpoint (actually any endpoint works ex. `/anything`). Just trying to make it more friendly to individuals not willing to dig into the source.

I do plan on adding documentation for discord notifications in the next week.  
 
I have tested this PR both with and without the /slack endpoint and can verify that it functions the same either way.

piksel:
> The lack of documentation when it comes to using discord notifications requires digging into the source to determine that the webhook URL needs an extra endpoint [...]

That's because the discord-over-slack-API is not recommended, nor supported. We only added the helper so that it wouldn't break the current users use-cases. 

>  I however have not looked through the shoutrrr source and would not be surprised if shoutrrr is adding the /slack endpoint back in.

No, it doesn't, because it's got native discord support. It's not using the slack-emulation in the discord API.

piksel (approved):
LGTM! Thanks!

PR #1319, opened 2022-07-12
docs: clarify container label usage

EDIflyer:
As discussed in #1318 - just to aid clarity for new users and also make docker-compose element clear.

### DIFF
--- a/docs/container-selection.md
+++ b/docs/container-selection.md
@@ -7,7 +7,7 @@ There are two options:
 
 ## Full Exclude 
 
-If you need to exclude some containers, set the _com.centurylinklabs.watchtower.enable_ label to `false`.
+If you need to exclude some containers, set the _com.centurylinklabs.watchtower.enable_ label to `false`.  For clarity this should be set **on the container(s)** you wish to be ignored, this is not set on watchtower.
 
 ```docker
 LABEL com.centurylinklabs.watchtower.enable="false"
@@ -19,7 +19,16 @@ Or, it can be specified as part of the `docker run` command line:
 docker run -d --label=com.centurylinklabs.watchtower.enable=false someimage
 ```
 
-If you need to [include only containers with the enable label](https://containrrr.github.io/watchtower/arguments/#filter_by_enable_label), pass the `--label-enable` flag or the `WATCHTOWER_LABEL_ENABLE` environment variable on startup and set the _com.centurylinklabs.watchtower.enable_ label with a value of `true` for the containers you want to watch.
+Or, in the docker-compose file:
+``` yaml
+services:
+  mycontainer:
+    [other fields such as container_name, entrypoint, environment, etc.]
+    labels:
+      - "com.centurylinklabs.watchtower.enable=false"
+```
+
+If instead you want to [only include containers with the enable label](https://containrrr.github.io/watchtower/arguments/#filter_by_enable_label), pass the `--label-enable` flag or the `WATCHTOWER_LABEL_ENABLE` environment variable on startup for watchtower and set the _com.centurylinklabs.watchtower.enable_ label with a value of `true` on the containers you want to watch.
 
 ```docker
 LABEL com.centurylinklabs.watchtower.enable="true"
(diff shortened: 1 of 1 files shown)


piksel:
You can make the examples tabbed, like on the [start page](https://containrrr.dev/watchtower/) like this:
https://github.com/containrrr/watchtower/blob/30f36b3ca29f0e1b11d0be0d4ac0900faceeaeb4/docs/index.md?plain=1#L47-L65

EDIflyer:
@piksel good suggestion - still getting my head round mkdocs - have tweaked to 3 tabs for the 2 scenarios, hopefully have got it right but obviously doesn't show up on Github's markdown interpreter!

EDIflyer:
sorry had renamed my branches so they made more sense, didn't realise that would automatically close the PR!

docs/container-selection.md
```diff
@@ -7,29 +7,57 @@ There are two options:
...
+    version: "3" 
+    services: 
+       watchtower: 
+         image: containrrr/watchtower 
+         volumes: 
+           - /var/run/docker.sock:/var/run/docker.sock
+        labels:
+          - "com.centurylinklabs.watchtower.enable=false"
+    ```
+
```
piksel: Ehm.... 😅

>  For clarity this should be set **on the container(s)** you wish to be ignored, this is not set on watchtower.

EDIflyer: Haha oops yeah good point 😂 I was trying to make my example really generic and stupidly picked watchtower as an example container 🤦🤦 I'll change to nginx!

EDIflyer: OK now changed to `someimage` as per other examples - thanks for spotting!

docs/container-selection.md
```diff
@@ -7,29 +7,67 @@ There are two options:
...
+    services:
+      someimage:
+        command:
+          - serve
+          - --dev-addr=0.0.0.0:8000
+        container_name: someimage
+        entrypoint:
+          - somepoint
+        environment:
+          - PATH=/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
piksel:
```suggestion
      someimage:
        container_name: someimage
```
I would suggest leaving out all the stuff that is not relevant here, especially since they have "real" values. Setting the `PATH` to `/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin` is not relevant to this feature, so it would be up to the user reading the docs to guess what matters and what doesn't.

docs/container-selection.md
```diff
@@ -7,29 +7,67 @@ There are two options:
...
```
piksel:
```suggestion
        container_name: someimage
```

Same as below. Only include the parts relevant for the feature.

docs/container-selection.md
```diff
@@ -7,29 +7,67 @@ There are two options:
...
-docker run -d --label=com.centurylinklabs.watchtower.enable=true someimage
-```
+If you need to exclude some containers, set the _com.centurylinklabs.watchtower.enable_ label to `false`.  For clarity this should be set **on the container(s)** you wish to be ignored, this is not set on watchtower.
+
+=== "dockerfile"
+
+    ```docker
+    LABEL com.centurylinklabs.watchtower.enable="false"
+    ```
+=== "dockerrun"
```
piksel:
```suggestion
=== "docker run"
```

docs/container-selection.md
```diff
@@ -7,29 +7,67 @@ There are two options:
...
+    ```
+
+If instead you want to [only include containers with the enable label](https://containrrr.github.io/watchtower/arguments/#filter_by_enable_label), pass the `--label-enable` flag or the `WATCHTOWER_LABEL_ENABLE` environment variable on startup for watchtower and set the _com.centurylinklabs.watchtower.enable_ label with a value of `true` on the containers you want to watch.
+
+=== "dockerfile"
+
+    ```docker
+    LABEL com.centurylinklabs.watchtower.enable="true"
+    ```
+=== "dockerrun"
```
piksel:
```suggestion
=== "docker run"
```

piksel (changes requested):
Looks good overall, but I think the examples should be reduced a bit to make it clearer what is actually required.

EDIflyer:
> Looks good overall, but I think the examples should be reduced a bit to make it clearer what is actually required.

Thanks - all very sensible suggestions 👍

piksel (approved):

PR #1337, opened 2022-08-01
feat: add porcelain output

### LINKED ISSUE #1336: Allow using stdout for session results to be used in scripts
**Quick intro**

Hi,
I realize that it might actually be the intended behavior and thus that it should not be considered as a bug. If that's the case, I'm sorry and this issue could be closed :smile:   
However, if that's the intended behavior, I think there could be a debate about it (that could lead to a feature request). I gave more information about that in the "additional context" section.  
Anyway, thanks for your awesome work !  

**Describe the bug**

The regular info messages shown below are written to standard error (stderr) instead of standard output (stdout) when using watchtower with the "run once" mode, even tho the command exits with a 0 return code (command used : `sudo docker run --rm -v /var/run/docker.sock:/var/run/docker.sock containrrr/watchtower --cleanup --run-once`) :
> time="2022-07-31T21:55:59Z" level=info msg="Watchtower 1.4.0"  
> time="2022-07-31T21:55:59Z" level=info msg="Using no notifications"  
> time="2022-07-31T21:55:59Z" level=info msg="Checking all containers (except explicitly disabled with label)"  
> time="2022-07-31T21:55:59Z" level=info msg="Running a one time update."  
> time="2022-07-31T21:56:02Z" level=info msg="Session done" Failed=0 Scanned=3 Updated=0 notify=no  
> time="2022-07-31T21:56:02Z" level=info msg="Waiting for the notification goroutine to finish" notify=no  

**To Reproduce**

Run the command and echo the return code shows 0 (which is expected) :  
`sudo docker run --rm -v /var/run/docker.sock:/var/run/docker.sock containrrr/watchtower --cleanup --run-once ; echo $?`  
> time="2022-07-31T21:59:47Z" level=info msg="Watchtower 1.4.0"  
> time="2022-07-31T21:59:47Z" level=info msg="Using no notifications"  
> time="2022-07-31T21:59:47Z" level=info msg="Checking all containers (except explicitly disabled with label)"  
> time="2022-07-31T21:59:47Z" level=info msg="Running a one time update."  
> time="2022-07-31T21:59:50Z" level=info msg="Session done" Failed=0 Scanned=3 Updated=0 notify=no  
> time="2022-07-31T21:59:50Z" level=info msg="Waiting for the notification goroutine to finish" notify=no  
> **0**  
  
However, redirecting standard output to "/dev/null" (or into a file) does nothing :  
`sudo docker run --rm -v /var/run/docker.sock:/var/run/docker.sock containrrr/watchtower --cleanup --run-once > /dev/null`
> time="2022-07-31T22:02:51Z" level=info msg="Watchtower 1.4.0"  
> time="2022-07-31T22:02:51Z" level=info msg="Using no notifications"  
> time="2022-07-31T22:02:51Z" level=info msg="Checking all containers (except explicitly disabled with label)"  
> time="2022-07-31T22:02:51Z" level=info msg="Running a one time update."  
> time="2022-07-31T22:02:54Z" level=info msg="Session done" Failed=0 Scanned=3 Updated=0 notify=no  
> time="2022-07-31T22:02:54Z" level=info msg="Waiting for the notification goroutine to finish" notify=no  
  
While redirecting standard error to "/dev/null" (or into a file) does redirect the ouput :
`sudo docker run --rm -v /var/run/docker.sock:/var/run/docker.sock containrrr/watchtower --cleanup --run-once 2> /dev/null` --> gives no output  
  
`sudo docker run --rm -v /var/run/docker.sock:/var/run/docker.sock containrrr/watchtower --cleanup --run-once 2> test.txt ; cat test.txt`
> time="2022-07-31T22:07:40Z" level=info msg="Watchtower 1.4.0"  
> time="2022-07-31T22:07:40Z" level=info msg="Using no notifications"  
> time="2022-07-31T22:07:40Z" level=info msg="Checking all containers (except explicitly disabled with label)"  
> time="2022-07-31T22:07:40Z" level=info msg="Running a one time update."  
> time="2022-07-31T22:07:43Z" level=info msg="Session done" Failed=0 Scanned=3 Updated=0 notify=no  
> time="2022-07-31T22:07:43Z" level=info msg="Waiting for the notification goroutine to finish" notify=no    
  
This behavior also happens if a container gets updated (at first I was wondering if those messages were designed to go to standard error if no update were available).

**Expected behavior**

I expected those info messages to be written in standard output and not standard error.
If those info messages are meant to be written in standard error, I may have a suggestion (see the "Additional context" section).

**Environment**

Tested on 2 different environment (same result for both)   
  
***Environment 1***
- Debian 11 installed directly on hardware
- X86_64
- Docker version 20.10.17, build 100c701

***Environment 2***
- Arch Linux VM installed on Proxmox
- X86_64
- Docker version 20.10.17, build 100c70180f

**Additional context**

As I said in the introduction, I initially thought it was a bug but then realized it could actually be the intended behavior (I'm sorry if that's the case).   
I understand that there's no real/standard convention about redirecting logging outputs to `stdout` or `stderr`, both are acceptable specially when a command doesn't have a "regular output" like it's the case here. 
However, these info messages contain a crucial information (at least to me) : The name of container that got updated.  
  
Here's my situation (to illustrate my point) :  
I'm not a fan of automatic updates processes. However, I like to simplify them as much as possible, (especially since I have a lot of servers that run docker containers). This is where the watchtower's `run once` mode becomes handy.  
So, to simplify my docker containers update process, I wrote an ansible playbook that launches watchtower with the `run once` mode on every of my servers, which I launch manually and regularly. 
To be able to see which containers got updated by watchtower (in order to know which ones need to be tested afterward to validate the update), I asked the playbook to redirect the standard output of the watchtower command to its own output. This is how I saw that the standard output of Watchtower does not contains info messages and thus the name of the container(s) that got updated.  
Of course, redirecting standard error of the Watchtower command instead of standard output in the ansible playbook (or redirecting `stderr` in `stdout` in the command) does work, but it feels a little weird to have to redirect **stderr** to get a standard information such as the name of containers.
  
According to that, here's my thought about this :
- If the info messages were not specifically meant to be written in the `stderr`, then I guess this bug report is legit.
- If the info messages were indeed meant to be written in the `stderr`, I would suggest a feature request that does output the name of the container(s) that got updated in the `stdout`. So people like me that use watchtower with the `run once` mode inside a script or an ansible playbook (for instance) doesn't have to redirect `stderr` to get this information (which is crucial for me). *One could argue that people using watchtower directly as a command on their machines (and not inside an ansible playbook like I do) would then get the name of updated container(s) twice as both `stdout` and `stderr` are shown by default when launching a command. But I still think that a dedicated line that gives the name of container(s) that got updated once treatments are finished would be useful and more easily readable then retrieving those names in the info logs messages in the form of `Stopping /container_name (container_id) with SIGTERM\`, `Creating /container_name\`, etc... Especially if you have servers that host a lot of docker containers.*

I know my case may be a bit particular but I think this would also benefits people that uses watchtower is a more "standard" way.
Anyway, as I also said earlier, **thanks a lot for your awesome work with Watchtower**, and sorry for this way too long report... :smile:

If this is indeed the intended behavior, don't hesitate to close this bug report and I'll open a feature request instead !

piksel:
This PR adds the `--porcelain` flag to help with using watchtower in scripts. When set, it uses a built-in template (`porcelain.v1.summary-no-log`) and writes notifications to `stdout`.
The output is meant to be stable and easily processed by common scripting tools like `awk` etc. 

Versioning of the templates is done by passing the version to `--porcelain` as an argument, with only `v1` being supported for now. That way we can add newer versions of templates without breaking scripts.

The output of the `v1` template looks like this:
```
container_name1 (container_image1:tag): Updated
container_name2 (container_image2:tag): Fresh Error: no credentials
```
(fresh here just means that it's the latest version _we know of_, but checking failed so it might not be)

or:
```
no containers matched filter
```

Fixes #1336

### DIFF
--- a/internal/flags/flags.go
+++ b/internal/flags/flags.go
@@ -510,12 +510,12 @@ func ProcessFlagAliases(flags *pflag.FlagSet) {
 		setFlagIfDefault(flags, `notification-template`, `porcelain.v1.summary-no-log`)
 	}
 
-	// update schedule flag to match interval if it's set
-	if flags.Changed(`interval`) {
-		schedule, _ := flags.GetString(`schedule`)
-		if len(schedule) > 0 {
-			log.Fatal(`Only schedule or interval can be defined, not both.`)
-		}
+	if flags.Changed(`interval`) && flags.Changed(`schedule`) {
+		log.Fatal(`Only schedule or interval can be defined, not both.`)
+	}
+
+	// update schedule flag to match interval if it's set, or to the default if none of them are
+	if flags.Changed(`interval`) || !flags.Changed(`schedule`) {
 		interval, _ := flags.GetInt(`interval`)
 		flags.Set(`schedule`, fmt.Sprintf(`@every %ds`, interval))
 	}
--- a/internal/flags/flags.go
+++ b/internal/flags/flags.go
@@ -551,6 +551,3 @@ func setFlagIfDefault(flags *pflag.FlagSet, name string, value string) {
 		log.Errorf(`Failed to set flag: %v`, err)
 	}
 }
- err)
-	}
-}
--- a/internal/flags/flags.go
+++ b/internal/flags/flags.go
@@ -2,6 +2,7 @@ package flags
 
 import (
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"strings"
@@ -152,22 +153,31 @@ func RegisterSystemFlags(rootCmd *cobra.Command) {
 		"",
 		viper.GetString("WATCHTOWER_HTTP_API_TOKEN"),
 		"Sets an authentication token to HTTP API requests.")
+
 	flags.BoolP(
 		"http-api-periodic-polls",
 		"",
 		viper.GetBool("WATCHTOWER_HTTP_API_PERIODIC_POLLS"),
 		"Also run periodic updates (specified with --interval and --schedule) if HTTP API is enabled")
+
 	// https://no-color.org/
 	flags.BoolP(
 		"no-color",
 		"",
 		viper.IsSet("NO_COLOR"),
 		"Disable ANSI color escape codes in log output")
+
 	flags.StringP(
 		"scope",
 		"",
 		viper.GetString("WATCHTOWER_SCOPE"),
 		"Defines a monitoring scope for the Watchtower instance.")
+
+	flags.BoolP(
+		"porcelain",
+		"P",
+		viper.GetBool("WATCHTOWER_PORCELAIN"),
+		"Write session results to stdout (alias for --notification-url logger:// --notification-log-stdout)")
 }
 
 // RegisterNotificationFlags that are used by watchtower to send notifications
@@ -342,6 +352,10 @@ Should only be used for testing.`)
 		viper.GetString("WATCHTOWER_WARN_ON_HEAD_FAILURE"),
 		"When to warn about HEAD pull requests failing. Possible values: always, auto or never")
 
+	flags.Bool(
+		"notification-log-stdout",
+		viper.GetBool("WATCHTOWER_NOTIFICATION_LOG_STDOUT"),
+		"Write notification logs to stdout instead of logging (to stderr)")
 }
 
 // SetDefaults provides default values for environment variables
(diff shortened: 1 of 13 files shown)


pkg/notifications/common_templates.go
```diff
@@ -0,0 +1,36 @@
...
+	  {{- end -}}
+    {{- end -}}
+  {{- end -}}
+{{- else -}}
+  {{range .Entries -}}{{.Message}}{{"\n"}}{{- end -}}
+{{- end -}}`,
+
+	`porcelain.v1.summary-no-log`: `
+{{- if .Report -}}
+{{len .Report.All}} containers matched filter{{println}}
```
piksel: This is probably not necessary and only makes it harder to filter the output.
Perhaps something should be emitted when there are no matching containers though, as no output could be caused by other factors.

internal/flags/flags.go
```diff
@@ -152,22 +153,31 @@ func RegisterSystemFlags(rootCmd *cobra.Command) {
...
 		"scope",
 		"",
 		viper.GetString("WATCHTOWER_SCOPE"),
 		"Defines a monitoring scope for the Watchtower instance.")
+
+	flags.BoolP(
+		"porcelain",
+		"P",
+		viper.GetBool("WATCHTOWER_PORCELAIN"),
+		"Write session results to stdout (alias for --notification-url logger:// --notification-log-stdout)")
```
```suggestion
	flags.StringP(
		"porcelain",
		"P",
		viper.GetString("WATCHTOWER_PORCELAIN"),
		"Write session results in a stable versioned format to stdout. Only possible argument is v1")
```
...or scrap versioning.

pkg/notifications/common_templates.go
piksel: Removed! Replaced with the message `no containers matched filters` (when applicable).

internal/flags/flags.go
piksel: Added the version argument.

PR #1345, opened 2022-08-17
feat: allow log level to be set to any level

matthewmcneely:
When using watch tower for a convenient way to reload freshly built images from the local Docker registry (with --no-pull and a poll interval of one second), the "Session done" log message spams the logs: `level=info msg="Session done" Failed=0 Scanned=3 Updated=0 notify=no`.

Since this is really debug-oriented messaging, seems appropriate.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -87,11 +87,11 @@ func PreRun(cmd *cobra.Command, _ []string) {
 		})
 	}
 
-	if enabled, _ := f.GetBool("debug"); enabled {
-		log.SetLevel(log.DebugLevel)
-	}
-	if enabled, _ := f.GetBool("trace"); enabled {
-		log.SetLevel(log.TraceLevel)
+	rawLogLevel, _ := f.GetString(`log-level`)
+	if logLevel, err := log.ParseLevel(rawLogLevel); err != nil {
+		log.Fatalf("Invalid log level: %s", err.Error())
+	} else {
+		log.SetLevel(logLevel)
 	}
 
 	scheduleSpec, _ = f.GetString("schedule")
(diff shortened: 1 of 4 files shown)


piksel:
Why do you consider the result of a session "debug" information? What is it that you are looking for in the logs if the summary of a run is "too much"?

matthewmcneely:
It's a fine line sure, I get your point. In my case I'm trying to use watchtower to check the local docker registry every second for new images. So the logs are getting full of "Session done" messages.

It would be fine if I could actually control the general watchtower logging level that's going to stderr (I'd just set it to WARN). Seems that I can change the notification logging level, but that does not effect general logging it seems, only the mail, slack etc notification system.

simonaronsson:
This is way too much of an edge case to warrant changing the defaults. Having it as INFO likely is what most users would expect.

matthewmcneely:
@simskij OK, understood.

Am I correct that there is no way (flag, env var) to configure the default logging? I understand I can change the *Notification* logging level that sends out emails, slack messages, etc.—but not the 'console' logging.

For my use case (checking local registry every second), that is what I first attempted to do.

simonaronsson:
> @simskij OK, understood.
> 
> Am I correct that there is no way (flag, env var) to configure the default logging? I understand I can change the _Notification_ logging level that sends out emails, slack messages, etc.—but not the 'console' logging.
> 
> For my use case (checking local registry every second), that is what I first attempted to do.

That is kind of correct. You can switch to tracing or debugging, but that really is the opposite of what you're looking for. I would suggest extending the command line arguments with a flag to set the level to a less permissive one, following the same pattern as in https://github.com/containrrr/watchtower/blob/146576818c5480ab76758455fc501e10fe5e3089/cmd/root.go#L90-L95. This should be a fairly trivial change only requiring you to make a couple of minor additions in a few files.

piksel:
Agreed. Using the new flag handling patterns, the `--debug` and `--trace` flags should be aliases for `--log-level debug` etc.

PR #1406, opened 2022-10-03
Added missing arguments to slack notification documentation

Choromanski:
Added the following two arguments to the notification documentation.

- notification-slack-icon-emoji
- notification-slack-icon-url


piksel:
The reason for not showing them is that it's only supported **sometimes** (using the legacy bot token). If we add this, we should add instructions for how to get the right type of token (if you can still get them from slack, since they are deprecated).

simonaronsson:
Thank you for the contribution, @Choromanski!

As @piksel stated, however, these are intentionally left out. To my knowledge you can't even generate legacy tokens anymore, so this would only work for older tokens (unless they have been revoked at this point?). Closing without merging.

PR #1412, opened 2022-10-05
docs: add dark mode

carueda:
This PR adds dark mode to the mkdocs-based documentation site. Pretty straightforward  with the great mkdocs tooling. Default mode will reflect user's OS preference, with ability to toggle dark/light mode after that. Main adjustment only in mkdocs.yml, but I also adjusted one style in theme.css so the links in dark mode are a bit more legible.

image

### DIFF
--- a/docs/stylesheets/theme.css
+++ b/docs/stylesheets/theme.css
@@ -6,6 +6,10 @@
     --md-accent-fg-color--transparent: #00334310;
 }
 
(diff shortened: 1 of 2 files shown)


piksel (approved):
Great idea! There really ought to be a containrrr-dark theme though, but it can be added in another PR.

piksel:
The codacy/csslint error seems to be just wrong. Ignoring it.

PR #1415, opened 2022-10-07
feat: add no-pull label for containers

### LINKED ISSUE #1256: Proposal: Specify no-pull per service/container via label
I am watching multiple containers with watchtower. One of them only gets local image updates (built via cron). It would be nice if watchtower could restart/update that container. But currently, I get this error message:

`Unable to update container "/name_name_1": Error response from daemon: pull access denied for custom/name, repository does not exist or may require 'docker login': denied: requested access to the resource is denied. Proceeding to next.`

It would be nice to be able to specify the no-pull option just for this container via a label.

nedzadalibegovic:
Resolves #1256.

This pull request adds support for a "no-pull" label (`com.centurylinklabs.watchtower.no-pull`) to enable the global functionality on a per-container basis.

### DIFF
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -276,7 +276,7 @@ func (client dockerClient) RenameContainer(c Container, newName string) error {
 func (client dockerClient) IsContainerStale(container Container) (stale bool, latestImage t.ImageID, err error) {
 	ctx := context.Background()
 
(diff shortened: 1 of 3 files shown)


piksel:
This looks good overall, but needs tests and a mention in the docs.

PR #1481, opened 2022-11-20
feat: ignore removal error due to non-existing containers

### LINKED ISSUE #1480: Compatibility issue with docker compose abort-on-container-exit flag
### Describe the bug

Watchtower causes an issue for me in combination with the `--abort-on-container-exit` flag for dockers `compose up` subcommand.

Docker compose removed the container on its own after the container was stopped. This leads to an error when watchtower attempts to remove the container.

`Error: No such container: 425173...`

The error is thrown by this [ContainerRemove](https://github.com/containrrr/watchtower/blob/d744c3488667d6cc4bb8258bb4b351bfd80e1602/pkg/container/client.go#L194) call.

### Steps to reproduce

1. Run container with `docker compose up --abort-on-container-exit`
2. Update image with watchtower

### Expected behavior

I expect for watchtower to be able to skip the container removal, either automatically if the container is already gone or with an explicit option.

### Screenshots

_No response_

### Environment

- Debian 11 amd64
- Docker 20.10.21

### Your logs

```text
...
Nov 20 23:09:37 docker[17234]: watchtower  | time="2022-11-20T23:09:37+01:00" level=debug msg="Found a match"
Nov 20 23:09:37 docker[17234]: watchtower  | time="2022-11-20T23:09:37+01:00" level=debug msg="No pull needed. Skipping image."
Nov 20 23:09:37 docker[17234]: watchtower  | time="2022-11-20T23:09:37+01:00" level=debug msg="No new images found for /traefik"
Nov 20 23:09:37 docker[17234]: watchtower  | time="2022-11-20T23:09:37+01:00" level=debug msg="This is the watchtower container /watchtower"
Nov 20 23:09:37 docker[17234]: watchtower  | time="2022-11-20T23:09:37+01:00" level=info msg="Stopping /website (425173d9c76c) with SIGTERM"
Nov 20 23:09:38 docker[17234]: watchtower  | time="2022-11-20T23:09:38+01:00" level=debug msg="Removing container 425173d9c76c"
Nov 20 23:09:38 docker[17234]: watchtower  | time="2022-11-20T23:09:38+01:00" level=error msg="Error: No such container: 425173d9c76c6f794674ac37f4999ce0817967ec6819c2ba21c61572bed4f383"
Nov 20 23:09:38 docker[17234]: watchtower  | time="2022-11-20T23:09:38+01:00" level=info msg="Session done" Failed=1 Scanned=5 Updated=0 notify=no
```

nothub:
This PR allows watchtower to ignore a failure to remove containers that doesn't exist. This can be useful when the start of the containers is managed by docker compose or an external system such as systemd.

Watchtower causes an issue for me in combination with the `--abort-on-container-exit` flag for dockers `compose up` subcommand.

Docker compose removed the container on its own after the container was stopped. This leads to an error when watchtower attempts to remove the container.

`Error: No such container: 425173...`

The error is thrown by this [ContainerRemove](https://github.com/containrrr/watchtower/blob/d744c3488667d6cc4bb8258bb4b351bfd80e1602/pkg/container/client.go#L194) call.

Fixes #1480

### DIFF
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -192,6 +192,10 @@ func (client dockerClient) StopContainer(c Container, timeout time.Duration) err
 		log.Debugf("Removing container %s", shortID)
 
 		if err := client.api.ContainerRemove(bg, idStr, types.ContainerRemoveOptions{Force: true, RemoveVolumes: client.RemoveVolumes}); err != nil {
+			if sdkClient.IsErrNotFound(err) {
+				log.Debugf("Container %s not found, skipping removal.", shortID)
+				return nil
+			}
 			return err
 		}
 	}
--- a/pkg/container/client_test.go
+++ b/pkg/container/client_test.go
@@ -1,6 +1,8 @@
 package container
 
 import (
+	"time"
+
 	"github.com/containrrr/watchtower/pkg/container/mocks"
 	"github.com/containrrr/watchtower/pkg/filters"
 	t "github.com/containrrr/watchtower/pkg/types"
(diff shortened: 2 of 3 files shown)


piksel:
Just skipping all container removal sounds a bit inflexible, and only fixes the problem if **all** containers are set to auto-remove.
It's probably better to ignore the error and try to create the new container anyway.  
Perhaps we can even detect that the container is set to auto-remove (should be the same as `docker run --rm`.

nothub:
Yes you are right, checking for each container is a better choice.
I pushed a change with that behaviour.
The error is now skipped if it is of type [ErrNotFound](https://pkg.go.dev/github.com/docker/docker/errdefs#ErrNotFound).

nothub:
Is there anything else you want me to change for the PR to be accepted?

piksel:
I mostly haven't had time. Sorry. I will get to reviewing it as soon as I get some spare time.

piksel:
Yeah, watchtower already checks for `AutoRemove`, but since this is just the docker-compose process manually stopping and removing the containers it's a special case. Ignoring that the container was not found when trying to remove it should be fine in general though.

piksel (approved):
This change seems to be the most sane course to take regardless of whether it's caused by the issue this PR attempts to fix.
I added a test for the "normal" case (container still existing after being stopped), as well as one for when the container could not be found.

nothub:
Awesome :) Thanks a lot for your work on this project!

piksel:
@all-contributors add @nothub for code

Issue #1485, opened 2022-11-29
Timeout and multiple bugs when using new notification url


kirbylink:
### Describe the bug

As discussed in #1374 i create a new issue for my problem.
I switched to the new notification url and among other things i get a time out when watchtower tries to send a notification after updating container.
When setting `WATCHTOWER_NO_STARTUP_MESSAGE=false` i will be notified, so the notification url seems to be correct.

Additional i found three more bugs:
- When starting Watchtower with `notify-upgrade` and legacy setup for smtp and gotify, the url for gotify may not have been converted correctly:
According to documentation the url should be like
`gotify://gotify.example.com:443/AzyoeNS.D4iJLVa/?title=Great+News&priority=1` but in my case Watchtower created an url like:
`gotify://gotify.example.com:443/AzyoeNS.D4iJLVa?title=Great+News&priority=1` (without `/` after the token)
If needed, i can restore the old setup and post the complete anonymized created url.
- Secondly i set `WATCHTOWER_NOTIFICATION_DELAY=300` but watchtower tries to send immediately after updating the container.
- Third the error message seems to fetch the wrong service for logging purpose (because the startup notification were handled correctly). For email the log says `service=gotify` and for gotify the log says `service=smtp`

### Steps to reproduce

1. Create docker-compose file like:
```yaml
version: "3"
services:
  watchtower:
    container_name: watchtower
    restart: unless-stopped
    image: containrrr/watchtower
    network_mode: bridge
    environment:
      - WATCHTOWER_MONITOR_ONLY=false
      - WATCHTOWER_SCHEDULE=@midnight
      - WATCHTOWER_NO_STARTUP_MESSAGE=false
      - WATCHTOWER_NOTIFICATION_URL=${WATCHTOWER_NOTIFICATION_URL}
      - WATCHTOWER_NOTIFICATIONS_HOSTNAME=${WATCHTOWER_NOTIFICATIONS_HOSTNAME}
      - WATCHTOWER_NOTIFICATION_DELAY=300
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      - /etc/localtime:/etc/localtime:ro
    labels:
      - "io.portainer.accesscontrol.users=${PORTAINER_USER}"
```
2. Create .env file like:
```
[...]

PR #1486, opened 2022-11-29
docs: fix env variable name `notification(s)-delay`

kirbylink:
Setting correct env. variable for --notifications-delay in notifications.md
See #1485


Tentoe (approved):
It's an S and it is correct.
How is it not merged yet?

piksel (approved):
Yeah, I don't know. Must've missed it.

PR #1501, opened 2022-12-16
Add podman support

### LINKED ISSUE #1060: [Feature Request] Support for Podman
Currently `watchtower` requires `/var/run/docker.sock`, which is not present in a system with only `podman` installed (i.e. where `docker` is not installed).

Potentially supporting `podman` in the future.

Thanks!

d-513:
Closes #1060

When making container, it checks if running in podman (by checking if /run/.containerenv file exists, that only is made when running with podman). If yes, then it applies workarounds:

- Removes leading slash from container name (podman doesn't support that)
- Doesnt set ulimits (also breaks podman)

### DIFF
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


d-513:
: error Not working yet

```
cannot set memory swappiness with cgroupv2: OCI runtime error
```

DocMAX:
> : error Not working yet
> 
> ```
> cannot set memory swappiness with cgroupv2: OCI runtime error
> ```

I get that error too now on multiple containers. How can i fix this now?

witten:
The ulimit thing may no longer be an issue in Podman 4.5.0+: https://github.com/containers/podman/pull/18078

0220tiezo1323 (changes requested):
No

adrianlzt:
Same error with podman version 5.2.2

PR #1526, opened 2023-01-17
fix(registry): image name parsing behavior

### LINKED ISSUE #1525: Cannot do a head request for private registry
### Describe the bug

I am having the exact same issue as described and discussed in https://github.com/containrrr/watchtower/issues/715, with the latest version of the image and the latest `registry:2` image.

My registry is located at `docker-registry.sub.domain.nl`, image names are single part, so `docker-registry.sub.domain.nl/imagename:latest`.

Watchtower tries `HEAD /v2/library/imagename/manifests/latest`. This fails, after which it "falls back to regular pull" and requests `HEAD /v2/imagename/manifests/latest`, which succeeds.

### Steps to reproduce

Have watchtower try to pull an image from a private registry.

### Expected behavior

`HEAD /v2/imagename/manifests/latest` -> no unnecessary pulls

### Screenshots

_No response_

### Environment

* Docker 20.10.11
* `containrrr/watchtower:latest` (latest version)
* `registry:2` (latest version)

### Your logs

```text
msg="Trying to load authentication credentials." container=/container image="docker-registry.sub.domain.nl/imagename:latest"
msg="Loaded auth credentials for user deploy, on registry docker-registry.sub.domain.nl/imagename:latest, from file /config.json"
msg="Got image name: docker-registry.sub.domain.nl/imagename:latest"
msg="Credentials loaded"
msg="Checking if pull is needed" container=/container image="docker-registry.sub.domain.nl/imagename:latest"
msg="Building challenge URL" URL="https://docker-registry.sub.domain.nl/v2/"
msg="Got response to challenge request" header="Basic realm=\"Registry Realm\"" status="401 Unauthorized"
msg="Parsing image ref" host=docker-registry.sub.domain.nl image=imagename normalized="docker-registry.sub.domain.nl/imagename:latest" tag=latest
msg="Doing a HEAD request to fetch a digest" url="https://docker-registry.sub.domain.nl/v2/library/login/manifests/latest"
msg="Could not do a head request for \"docker-registry.sub.domain.nl/imagename:latest\", falling back to regular pull." container=/container image="docker-registry.sub.domain.nl/imagename:latest"
msg="Reason: registry responded to head request with \"404 Not Found\", auth: \"not present\"" container=/container image="docker-registry.sub.domain.nl/imagename:latest"
msg="Pulling image" container=/container image="docker-registry.sub.domain.nl/imagename:latest"
msg="No new images found for /container"
```

### Additional context

_No response_

Pwuts:
## What this PR contributes
It replaces Watchtower's own image ref manipulation logic (wherever possible) by docker's own implementations as imported from [`docker/distribution/reference`](https://pkg.go.dev/github.com/docker/distribution@v2.8.1+incompatible/reference).

As a result, this PR:
* solves failing HEAD requests for images with a single-part path
* solves other off-spec behavior in some cases (such as references where the registry host has a port)
* reduces the amount of code to be maintained in this repository
* reduces the complexity of some parts of the code
* resolves [this](https://github.com/containrrr/watchtower/blob/a4d00bfd7589d45c8e3bf42d8d6d2dc0bbf133d4/pkg/registry/trust_test.go#L42-L48) todo
* Closes #1525

Changes are limited to *pkg/registry*, and deprecate *pkg/registry/helpers*.

<details>
<summary>How/what/why this works</summary>

In [reference/normalize.go], `ParseNormalizedNamed` and `ParseDockerRef` call `splitDockerDomain` which normalizes the registry domain from *index.docker.io* to *docker.io* if applicable. This allows us to directly use the properties of the `Named`/`NamedTagged`/`NamedDigested` that is returned by these (and other) functions.
* `helpers.NormalizeRegistry(normalizedName.String())` -> `reference.Domain(normalizedName)`
* `ExtractImageAndTag([...])` -> `reference.Path(normalizedRef), reference.Tag(normalizedRef)` (if container image is not pinned)
* `auth.GetScopeFromImageName(img, host)` -> `reference.Path(normalizedName)`
* etc...
</details>

[reference/normalize.go]: https://github.com/distribution/distribution/blob/362910506bc213e9bfc3e3e8999e0cfc757d34ba/reference/normalize.go

### Notable choices made along the way
* Defaulting to `docker.io` credentials (stored in the config under `index.docker.io` or `https://index.docker.io/v1/`) when an image name does not contain a registry host. This was an outstanding [todo](https://github.com/containrrr/watchtower/blob/a4d00bfd7589d45c8e3bf42d8d6d2dc0bbf133d4/pkg/registry/trust_test.go#L42-L48), and can be done safely [under the right conditions](#issuecomment-1386874300).
* Using `index.docker.io` to specify an image on Docker Hub is legacy, so I have replaced `index.docker.io` with `docker.io` in all relevant places, or removed it altogether, as with this PR, Watchtower should use the right Docker Hub domain by default when no registry is specified in an image name.
* `registry-1.docker.io` is only supposed to be used in setting up a pull-through cache registry, not for pulling images directly. Hence, I think it's reasonable to drop ["special support"](https://github.com/containrrr/watchtower/blob/a4d00bfd7589d45c8e3bf42d8d6d2dc0bbf133d4/pkg/registry/helpers/helpers.go#L28) for this case (and disable the corresponding test).
If anyone really wants to use that domain, they can still specify the image as `registry-1.docker.io/library/image`, just like they already need to do when using a docker hub mirror.
* Disabling [this test](https://github.com/containrrr/watchtower/blob/a4d00bfd7589d45c8e3bf42d8d6d2dc0bbf133d4/pkg/registry/auth/auth_test.go#L117), which tests for something that should never occur: a container with a familiar name (e.g. `ubuntu` for `docker.io/library/ubuntu`) that is not hosted on Docker Hub. Containers pulled from registries other than Docker Hub will always have the registry domain in `container.ImageName()`.

## Tests
After my updates, all tests succeed except for the two mentioned above, which I have commented for now.

I tried to improve test coverage as best as possible for the files that I changed, including a test in [pkg/registry/manifest/manifest_test.go](https://github.com/Pwuts/watchtower/blob/fix/ref-processing/pkg/registry/manifest/manifest_test.go#L65) for the specific issue that triggered me to do this work :)

## Updates to the documentation
As mentioned above, I have replaced `index.docker.io` by `docker.io` wherever relevant and appropriate.
Also added a heads-up + examples of using private registries on local and remote hosts.


piksel:
> Finally, I'm not sure if [WarnOnAPIConsumption](https://github.com/containrrr/watchtower/blob/a4d00bfd7589d45c8e3bf42d8d6d2dc0bbf133d4/pkg/registry/registry.go#L42-L59) will still be necessary after these modifications.

I'm not sure why that would be the case. Watchtower warns the user that the `HEAD` request failed, and `docker pull` is used instead. But since this might not be supported on all repositories, we only warn for those that we know support it (when set to `Auto`), to avoid spamming the logs (as most users don't have any way to fix it, nor gain that much benefit from the HEAD requests since they probably aren't rate-limited).

piksel:
> index.docker.io is legacy and no longer the primary domain for the Docker Hub registry, so I have swapped index.docker.io and docker.io in all relevant places

Unfortunately, I think this will break users configs, since they have been explicitly told to use the full `index.docker.io` domain in their `docker.config`s. But I think we can make a backwards compatibility fix for this...

Pwuts:
> Unfortunately, I think this will break users configs

Good catch! I'll try to push a patch today.

Pwuts:
> Unfortunately, I think this will break users configs

Users who have followed the instructions and specified the registry domain explicitly in their docker-compose config or `docker run` command should not run into problems, because `GetToken` eventually calls `reference.Parse()`, not `reference.ParseNormalizedNamed()`, meaning any specified domain will not be normalized or altered before using it to fetch credentials from the config file.

However, I see there's a [todo](https://github.com/containrrr/watchtower/blob/a4d00bfd7589d45c8e3bf42d8d6d2dc0bbf133d4/pkg/registry/trust_test.go#L42-L48) (from 2019) to elaborate `ParseServerAddress()` so that it assumes `docker.io` when no domain or registry host is specified. Is this still a todo or desired behavior? Because currently there is a test that specifies "parse server address_ should return the organization part if passed an image name missing server name". They can't both be true :)

~_Edit: I think it's a bad idea to have any kind of fallback behavior with regards to credentials when there is any uncertainty about the registry host, because falling back to credentials for `docker.io` means submitting credentials for `docker.io` to any registry for which credentials are not set. This is a security and credential leak risk -> very undesirable._~

**Edit 2:** [Apparently](https://github.com/distribution/distribution/blob/a4cec8ca82ae277aaa412f4e55bbc013f6d0d8e4/reference/normalize.go#L128), docker's native behavior is to only parse hosts that satisfy `strings.ContainsAny(host, ".:") || host == 'localhost'`, meaning it *is* always possible to distinguish between an organization name and a host.

piksel:
Yeah, the fallback to `docker.io` creds should only apply when the registry hostname is updated to `docker.io` (which is the normal case for when the ref lacks any repository host).

Pwuts:
Implemented in 0671ee8acc7d1a897ac9994b5e7223c03b77ef8d ([here](https://github.com/Pwuts/watchtower/blob/0671ee8acc7d1a897ac9994b5e7223c03b77ef8d/pkg/registry/trust.go#L78-L93)).

Pwuts:
0671ee8acc7d1a897ac9994b5e7223c03b77ef8d also improves test coverage

Pwuts:
@piksel could you approve another codecov run? I think it would be more positive this time. :)

Also, I'm not sure how to proceed. I have made all the semantic changes I intended, but now there's a certain amount of dead code. Should I wait with removing that until after a first round of CR?

Pwuts:
Thanks! Looks good, it's only complaining about `auth.go` because the test for `GetToken` is skipped for lack of GHCR credentials, and the coverage diff for `manifest.go` should be positive after cleaning up the now-dead code.

The security vulnerability ("password logging") that the CodeQL check warns for was already there when I got here, and is only in a `Tracef` call ¯\\\_(ツ)_/¯

Pwuts:
There we go:
* `GetRegistryAddress` gives `index.docker.io` for images specified with registry `docker.io`, `index.docker.io`, or without a registry domain
* ~`EncodedConfigAuth` will also try `credStore.Get("https://index.docker.io/v1/")` if `index.docker.io` is not found in the config file, since `docker login` without saves the credentials under that key by default~
* Updated the docs with a heads-up + examples for using local registry hosts

I think that was the last to be done here without further expanding the scope of this PR. :)

Pwuts:
@simskij can I ask for your judgment on the semantic changes so far? :)

piksel:
> The security vulnerability ("password logging") that the CodeQL check warns for was already there when I got here, and is only in a Tracef call ¯\_(ツ)_/¯

Yeah, but there really isn't any need for it anymore. I created a PR (#1534) to remove all the logging that has no clear usage scenario and could potentially leak credentials.

docs/private-registries.md
```diff
@@ -23,19 +23,23 @@ password `auth` string:
...
-    you would run it as
-    ```
-    docker run -d index.docker.io/myuser/myimage
-    ```
-
+(e.g., `my-private-registry.example.org`).
+
+!!! info "Using private images on Docker Hub"
+    To access private repositories on Docker Hub,
+    `<REGISTRY_NAME>` should be `index.docker.io`.
```
piksel:
```suggestion
    `<REGISTRY_NAME>` should be `docker.io`.
```
If using `docker.io` is fully supported, we might as well put it in the docs for new users, no?

Pwuts: Agreed, but it is a bit more nuanced than that. Docker uses `docker.io` internally as the default registry domain, but the Docker Registry API for Docker Hub is hosted on `index.docker.io`, and so docker libraries expect `config.json` to contain credentials for `index.docker.io`.

Worse, the Docker CLI does not recognize any other key than `https://index.docker.io/v1/`, and `docker/cli/cli/config/credentials` (that supplies `credStore.Get()`) applies some backwards compatibility magic to match this key when requesting `credStore.Get('index.docker.io')`. A call to `docker pull PRIVATE_IMAGE` (with `PRIVATE_IMAGE` an image on Docker Hub) fails with `config.json` only containing an entry for `index.docker.io`.

Pwuts: I have amended my last 2 commits to mimic this behavior in Watchtower and reflect it in the docs. For backwards compatibility, Watchtower will still recognize the `index.docker.io` credential entry, but also try `https://index.docker.io/v1/` if that doesn't exist.

Pwuts: I have amended my last 2 commits to "comply" with Docker's own behavior and reflect this in the documentation.

It took a few iterations to get it right and I didn't want to pollute the git history, hence the multiple force-pushes.

Pwuts: Note: the CLI [treats `index.docker.io` as an exception](https://github.com/docker/cli/blob/65438e008c20a6125a1319eca07dcc3d7d4e38eb/cli/command/registry.go#L73-L76) to the rule of how credentials are stored in `config.json`, and [a comment in `docker/cli`](https://github.com/docker/cli/blob/850fb6921cf7a208fd47d83074f27d0eb7ae268e/cli/config/credentials/file_store.go#L36-L42) remarks that this is legacy.

Would you agree that we should play it safe and ensure that a `config.json` manually generated according to the Watchtower docs should also work with the Docker CLI, even though this means we [tell users to go along with this legacy, inconsistent behavior](https://github.com/containrrr/watchtower/blob/6a6c7aec21ae35ad98923e65f72c8a277a6fd3d6/docs/private-registries.md?plain=1#L28-L36)?

Pwuts:
Repackaged the changes to reduce diff with upstream and make it easier to CR (per commit)

Pwuts:
@piksel any timeline for review/merge?

piksel:
No, sorry. But I will take a look as soon as I get some free time.

piksel (approved):

Pwuts:
Thanks a lot! :)

PR #1645, opened 2023-05-06
docs: add "HTTP API Mode" link to nav menu

SergeAx:
It would be nice to have it there and not look for a link in the "Arguments" section every time


piksel (approved):

lonix1:
@SergeAx I also stumbled on this and was about to submit a PR! :smile:

Just FYI the docs site does not show your change.

piksel:
Yeah, we have not made a release that includes this change yet. Should be pretty soon though!

PR #1669, opened 2023-06-11
Added container labels to ContainerReport interface

otaconix:
This PR aims to get labels from the containers Watchtower handles into the `ContainerReport` structure. I personally had an itch that needed scratching: I'm running a docker-compose stack, and I want to be able to see at a glance which projects and services need updating (I run Watchtower in _monitor-only_ mode).

With the changes in this PR, I can use the following notification template:

```go
{{- define "Name" -}}
  {{- if index .Labels "com.docker.compose.project" -}}
    {{- index .Labels "com.docker.compose.project" -}}
    :
    {{- index .Labels "com.docker.compose.service" -}}
  {{- else -}}
    {{- .Name -}}
  {{- end -}}
{{- end -}}

{{- with .Report -}}
  {{- if .Stale -}}
The following containers can be updated:
    {{- range .Stale }}
  - {{ template "Name" . }} (image: {{ .ImageName }})
    {{- end -}}
  {{- end -}}
  {{- if (or .Skipped .Failed) }}
The following errors occurred:
    {{- range .Skipped }}
  - {{ template "Name" . }}: {{ .Error }}
    {{- end -}}
    {{- range .Failed }}
  - {{ template "Name" . }}: {{ .Error }}
    {{- end -}}
  {{- end -}}
{{- end -}}
```

NB: I wouldn't mind adding a test to demonstrate the correct functioning of the code in this PR, but I wasn't sure what a good place would be to do that. Feedback welcome!


otaconix:
PS: I realize I could've used the container name in my notification template, but I personally prefer to format it myself by using the labels compose adds to the containers. 

If this proves to not be useful to anyone but me, I suppose this PR is not very useful.

0220tiezo1323 (changes requested):
No

PR #1690, opened 2023-07-08
Add a Matrix example to notifications.md

Krafting:
Why: I spent a few hours to get it setup right as I didn't understand what shoutrr was.

This might help others as well in the future. Example: https://lemmy.world/comment/928893

### DIFF
--- a/docs/notifications.md
+++ b/docs/notifications.md
@@ -49,8 +49,7 @@ Simple templates are used unless the `notification-report` flag is specified:
 
 To send notifications via matrix, the following command-line options, or their corresponding environment variables, should be set:
 
--   `--notification-url` (env. `WATCHTOWER_NOTIFICATION_URL`): The matrix service URL to be used. 
--   `--notification-template` (env. `WATCHTOWER_NOTIFICATION_TEMPLATE`): The format to be used for notifications.
+-   `--notification-url` (env. `WATCHTOWER_NOTIFICATION_URL`): The matrix service URL to be used.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


GridexX:
Indeed it could helps, :+1: For the contribution

piksel:
Hm, it doesn't seem like that template is anything matrix-specific? I think the main issue here is that it's not obvious from the docs that matrix support is built into watchtower (through shoutrrr). I guess it's still the legacy notification options that give that impression...

Krafting:
> Hm, it doesn't seem like that template is anything matrix-specific? I think the main issue here is that it's not obvious from the docs that matrix support is built into watchtower (through shoutrrr). I guess it's still the legacy notification options that give that impression...

It's indeed a bit confusing.

Also, the WATCHTOWER_NOTIFICATION_URL for matrix can be very weird, as you have the server url, the room id (which also contains the server URL) I had to do multiple attempt before finding the correct format, I just think this could really help others as well.

piksel:
The problem is that we should either include docs for all services, or none. Just having a section for matrix only makes the problem worse for the other services. We could have a list of supported services (which each link to the shoutrrr service docs). That would help with searchability as well. 
The server for the rooms is not required if it's the same as the "host" server (if I recall correctly). But otherwise I think any improvements to the docs should be redirected to shoutrrr (which I happen to be maintaining as well 😁).

simonaronsson:
I agree. We can definitely make improvements to https://github.com/canonical/shoutrrr, but I don't think it belongs here in watchtower.

PR #1699, opened 2023-07-24
fix: empty out the aliases on recreation

### LINKED ISSUE #521: too many network aliases when recreating the container
image

simonaronsson:
This PR makes sure old network aliases are truncated as containers get recreated.

resolves #521


piksel (approved):
I doubt we would need to provide any aliases in the config when creating, so clearing it should be fine. There should be a test of course, though 😁

piksel (approved):
LGTM!

PR #1725, opened 2023-08-09
feat: add --health-check command line switch

### LINKED ISSUE #1257: Add healthcheck command
No problem

Add a healthcheck command to the image.

**Describe alternatives you've considered**
reopen #972 :)

**Additional context**
maybe something like that :

healthcheck.sh
```bash
#!/bin/sh

PERIOD_12H=43200

MAX_DELAY=$(( PERIOD_12H + 60 ))

# delay since last loop (seconds)
DELAY=$(($(date +%s) - $(date +%s -r /tmp/last_loop)))

[ "${DELAY}" -gt "${MAX_DELAY}" ] && exit 1

exit 0
```

And for the dockerfile :

`HEALTHCHECK --interval=5m --timeout=2s --start-period=30s CMD sh healthcheck.sh`

it's just an example to improve and adjust with env var. The watchtower job has just to make a touch on the file to update it

bugficks:
This PR adds docker health check. Closes #1257


cmd/root.go
```diff
@@ -152,6 +152,15 @@ func Run(c *cobra.Command, names []string) {
 	enableMetricsAPI, _ := c.PersistentFlags().GetBool("http-api-metrics")
 	unblockHTTPAPI, _ := c.PersistentFlags().GetBool("http-api-periodic-polls")
 	apiToken, _ := c.PersistentFlags().GetString("http-api-token")
+	healthCheck, _ := c.PersistentFlags().GetBool("health-check")
+
+	if healthCheck {
+		// health check should not have pid 1
+		if os.Getpid() == 1 {
+			os.Exit(1)
```
piksel: If someone accidentally adds this to the command line, we should probably tell them the problem and prevent `restart: always` from pinning the CPU.

```suggestion
			time.Sleep(1 * time.Second)
			log.Fatal("The health check flag should never be passed to the main watchtower container process")
```

It would of course seem like this would also trigger if the container is up, but the main watchtower process has died, but this can never happen since that would stop the container.

docs/arguments.md
```diff
@@ -363,4 +363,15 @@ requests and may rate limit pull requests (mainly docker.io).
 Environment Variable: WATCHTOWER_WARN_ON_HEAD_FAILURE
      Possible values: always, auto, never
              Default: auto
-```
\ No newline at end of file
+```
+
+## Health check
+
+Run health check and exit.
```
piksel:
```suggestion
Returns a success exit code to enable usage with docker `HEALTHCHECK`. This check is naive and only returns checks whether there is another process running inside the container, as it is the only known form of failure state for watchtowers container.

!!! note "Only for HEALTHCHECK use"
    Never put this on the main container executable command line as it is only meant to be run from docker HEALTHCHECK.
```

docs/arguments.md
```diff
@@ -363,4 +363,15 @@ requests and may rate limit pull requests (mainly docker.io).
...
+
+```text
+            Argument: --health-check
+Environment Variable: WATCHTOWER_HEALTH_CHECK
+     Possible values: -
+             Default: -
```
piksel: This should never be set using an environment variable.

```suggestion
```

internal/flags/flags.go
```diff
@@ -186,6 +186,12 @@ func RegisterSystemFlags(rootCmd *cobra.Command) {
 		"log-level",
 		viper.GetString("WATCHTOWER_LOG_LEVEL"),
 		"The maximum log level that will be written to STDERR. Possible values: panic, fatal, error, warn, info, debug or trace")
+
+	flags.BoolP(
+		"health-check",
+		"",
+		viper.IsSet("WATCHTOWER_HEALTH_CHECK"),
```
```suggestion
		false,
```

piksel (changes requested):
Thanks for helping out! I think we should help make this less of a foot-gun for the users though.

piksel (approved):

PR #1726, opened 2023-08-11
fix(notifications): correctly set the delay from options

Tentoe:
`--notifications-delay` was ignored by soutrrr notifications because it wasn't set on creation.

I tested with `dockerfiles/Dockerfile.self-contained`

--- 

Caused by #1377 
Probably solves #1485 (needs confirmation)


piksel (approved):
Indeed. This even shows up as a warning in goland, I wonder why Codacy didn't catch this... Thanks for spotting it!

PR #1727, opened 2023-08-12
test(flags): ensure temp files are cleaned up

piksel:
Before, the tests actually left temporary files on windows due to floating file handles. This fix will fix warnings in `flags` and `flags_test` and ensures that the temporary files are removed (or it will fail the tests).


piksel:
Unfortunately, this PR lowers the test coverage, but not really. The same paths were just as untested before (they just logged the error and panicked or ignored the them). Additionally, the untested lines are pretty hard to simulate (I/O errors), or could never happen unless the actual code is incorrect (a flag that has secret file lookup enabled, but doesn't exist in the flags).

PR #1738, opened 2023-08-21
docs: clarify what volumes are removed when requested

szaimen:
Update docs to reflect what the `--remove-volumes` option actually does.

I tracked the code back to https://github.com/moby/moby/blob/a01bcf9767c91c91bde4be5a32b151e256e74523/client/container_remove.go#L13 and then searched in the api docs: https://docs.docker.com/engine/api/v1.43/#tag/Container/operation/ContainerDelete. There it is clearly stated: 
image

### DIFF
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -85,8 +85,8 @@ Environment Variable: WATCHTOWER_CLEANUP
              Default: false
 ```
 
-## Remove attached volumes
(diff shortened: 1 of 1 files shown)


szaimen:
Hi @piksel can you please review this? Thank you!

piksel (approved):

PR #1749, opened 2023-09-04
docs: bump go version in credential helper example

aioue:
Amazon credential helper updates mean using GO 1.16 will cause slice errors. e.g.

```
undefined: unsafe.Slice
```

[Using 1.17 as the source FROM fixes this](https://stackoverflow.com/questions/73735732/the-command-bin-sh-c-go-build-o-image-srv-returned-a-non-zero-code-2).


piksel (approved):
Thanks for contributing your solution!

PR #1753, opened 2023-09-06
ft: allow detect changes local images

### LINKED ISSUE #416: [Question] Can watchtower detect images changes in local repo?
Hello,

I was looking for ways to streamline docker images updates and came accross watchtower. After reading a bit of the documentation, I've come to know that it is possible for watchtower to pull images from local docker repositories. 

My question is, is it mandatory to first push an update to the local repository (or dockerhub for that  matter) in order for watchtower to detect the update and pull? If I were run a `docker pull` or `docker load` and download a new repo manually, will it be enough for watchtower to restart the exising containers with the new image?

Scenario:

1. I have custom_img:1.0 in my local machine.
2. There is a container running off of the above image. (tag: custom_img_container)
3. I manually push a new image (custom_img:2.0) in my local machine
Question: Will watchtower detect that there is now a new image and restart `custom_img_container` based off custom_img:2.0 ?

Thanks.

9bany:
fixes: https://github.com/containrrr/watchtower/issues/416

Desc: Allow watchtower detects changes local images, this is helpful if you run watchtower on your machine and you want to test your images on locally before release to production env.

How to use this feature:
- You need to add a label `com.centurylinklabs.watchtower.local-image=true` when run your container.
- The new image have the same name with the image used in your container.

### DIFF
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -311,6 +311,10 @@ func (client dockerClient) RenameContainer(c t.Container, newName string) error 
 func (client dockerClient) IsContainerStale(container t.Container) (stale bool, latestImage t.ImageID, err error) {
 	ctx := context.Background()
 
+	if isLocalImage, checkErr := container.LocalImage(); checkErr && isLocalImage {
+		return client.HasNewImage(ctx, container)
+	}
+
 	if !client.PullImages || container.IsNoPull() {
 		log.Debugf("Skipping image pull.")
 	} else if err := client.PullImage(ctx, container); err != nil {
--- a/pkg/container/container.go
+++ b/pkg/container/container.go
@@ -113,6 +113,22 @@ func (c Container) ImageName() string {
 	return imageName
 }
 
+// Local returns the value of the container local label and if the label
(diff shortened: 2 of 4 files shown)


piksel:
Hm, it seems like this would be equal to the `--no-pull` option, but for individual containers. Perhaps the label ought to have that name (`com.centurylinklabs.watchtower.no-pull`) and the a note in the docs about how it can be used for local images.

9bany:
Oh yeah, i think `com.centurylinklabs.watchtower.no-pull` label can work.

Let's me test it, and will close this pull if it's necessary.

Thank for your supports @piksel

piksel:
Haha, I suggested adding it, but you are right, it's already implemented. 

The point of the docs still stands though. I didn't realize that watchtower could even work with local-only images, but it makes sense (...and I couldn't find the label being mentioned when I looked for it).

jebabin:
I think this PR can be closed as the label no-pull exist and is documented (since march) in main

9bany:
i close this pull, thanks guy

PR #1754, opened 2023-09-07
feat: add a label take precedence argument

### LINKED ISSUE #1745: Update only specific containers

I'd like to be able to globally disable containers update but enable update for a specific container.

Currently, it's possible to disable globally containers update with WATCHTOWER_MONITOR_ONLY
But, unless i missed something, when WATCHTOWER_MONITOR_ONLY is set to true, there is no way to enable update for a specific container

The only way would be to set com.centurylinklabs.watchtower.monitor-only to true on all containers except the one that need to be update.  

Instead of having update enable by default then disable update on some containers, I'd like to have update disable by default then enable update on some contrainers

set com.centurylinklabs.watchtower.monitor-only to true on all containers except the one that need to be update.  

### Additional context

May be in case WATCHTOWER_MONITOR_ONLY=true && com.centurylinklabs.watchtower.monitor-only=false the update could be performed ?
It would imply to default com.centurylinklabs.watchtower.monitor-only to true when WATCHTOWER_MONITOR_ONLY=true

May be the following changes would do this
in update.go (line 37), changing
`shouldUpdate := stale && !params.NoRestart && !params.MonitorOnly && !targetContainer.IsMonitorOnly()`
to something like 
`shouldUpdate := stale && !params.NoRestart && !targetContainer.IsMonitorOnly()`
in update.go (line 75), removing
`if !params.MonitorOnly {`

in container.go (line 134), changing
```
func (c Container) IsMonitorOnly() bool {
	rawBool, ok := c.getLabelValue(monitorOnlyLabel)
	if !ok {
		return false
	}

	parsedBool, err := strconv.ParseBool(rawBool)
	if err != nil {
		return false
	}

	return parsedBool
}

```
to
```

func (c Container) IsMonitorOnly() bool {
	rawBool, ok := c.getLabelValue(monitorOnlyLabel)
	if !ok {
		if params.MonitorOnly {
			return true
		} else {
			return false
		}
	}

	parsedBool, err := strconv.ParseBool(rawBool)
	if err != nil {
		if params.MonitorOnly {
			return true
		} else {
			return false
		}
	}

	return parsedBool
}
```
but params is not defined in containers.go

jebabin:
This PR add an argument to have labels take precedence over arguments.
By default watchtower have argument taking precedence over label which is not very convenient because arguments are global while label specific to container.
This is useful to allow specific container to be updated when monitor only is set to true
it close #1745
Test has been added to ensure that container are updated when monitor only is global set to true but set to false on the container level.
Documentation has been updated to include this new argument.
Additionally documentation has been updated to better describe the label enable argument which not only disable update but also disable monitoring

### DIFF
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -205,7 +205,7 @@ Environment Variable: WATCHTOWER_POLL_INTERVAL
 ```
 
 ## Filter by enable label
-Update containers that have a `com.centurylinklabs.watchtower.enable` label set to true.
+Monitor and update containers that have a `com.centurylinklabs.watchtower.enable` label set to true.
 
 ```text
             Argument: --label-enable
@@ -215,7 +215,7 @@ Environment Variable: WATCHTOWER_LABEL_ENABLE
 ```
 
 ## Filter by disable label
-__Do not__ update containers that have `com.centurylinklabs.watchtower.enable` label set to false and 
+__Do not__ Monitor and update containers that have `com.centurylinklabs.watchtower.enable` label set to false and 
 no `--label-enable` argument is passed. Note that only one or the other (targeting by enable label) can be 
 used at the same time to target containers.
 
@@ -237,6 +237,18 @@ Environment Variable: WATCHTOWER_MONITOR_ONLY
 ```
 
 Note that monitor-only can also be specified on a per-container basis with the `com.centurylinklabs.watchtower.monitor-only` label set on those containers.
+
+## With label taking precedence over environment variables
+
+By default, environement variables will take precedence over labels. This means that if you set `WATCHTOWER_MONITOR_ONLY` to true, a container with `com.centurylinklabs.watchtower.monitor-only` set to false will not be updated. If you set `WATCHTOWER_LABEL_TAKE_PRECEDENCE` to true, then the container will also be updated
+
+
+```text
+            Argument: --label-take-precedence
+Environment Variable: WATCHTOWER_LABEL_TAKE_PRECEDENCE
+                Type: Boolean
+             Default: false
+```
 
 ## Without restarting containers
 Do not restart containers after updating. This option can be useful when the start of the containers
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -238,9 +238,12 @@ Environment Variable: WATCHTOWER_MONITOR_ONLY
 
 Note that monitor-only can also be specified on a per-container basis with the `com.centurylinklabs.watchtower.monitor-only` label set on those containers.
 
+See [With label taking precedence over arguments](##With label taking precedence over arguments) for behavior when both agument and label are set
+
+
 ## With label taking precedence over arguments
 
-By default, arguments will take precedence over labels. This means that if you set `WATCHTOWER_MONITOR_ONLY` to true or use `--monitor-only`, a container with `com.centurylinklabs.watchtower.monitor-only` set to false will not be updated. If you set `WATCHTOWER_LABEL_TAKE_PRECEDENCE` to true or use `--label-take-precedence`, then the container will also be updated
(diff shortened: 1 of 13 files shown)


internal/actions/update.go
```diff
@@ -34,7 +34,7 @@ func Update(client container.Client, params types.UpdateParams) (types.Report, e
 
 	for i, targetContainer := range containers {
 		stale, newestImage, err := client.IsContainerStale(targetContainer)
-		shouldUpdate := stale && !params.NoRestart && !params.MonitorOnly && !targetContainer.IsMonitorOnly()
+		shouldUpdate := stale && !params.NoRestart && (( !params.MonitorOnly && !targetContainer.IsMonitorOnly() ) || ( params.LabelPrecedence && !targetContainer.IsMonitorOnly() ))
```
piksel: I hated this line before and this makes the problem even worse... We have to be able to do this in a more legible way...

docs/arguments.md
```diff
@@ -238,6 +238,17 @@ Environment Variable: WATCHTOWER_MONITOR_ONLY
 
 Note that monitor-only can also be specified on a per-container basis with the `com.centurylinklabs.watchtower.monitor-only` label set on those containers.
 
+## With label taking precedence over environment variables
```
piksel: Environment variables sounds a bit too specific, as it concerns both command arguments and environmental variables (where the term arguments are generally preferred, since that is what we call the section).

internal/actions/update.go
piksel: Perhaps we can have a `containerMonitorOnly` that just deals with resolving whether a container has `monitor-only` enabled or not?

piksel (changes requested):
Thanks for helping out with this!

docs/arguments.md
jebabin: fixed

internal/actions/update.go
piksel: Okay, my suggestion is this: 
In the `IsEnabled` and `IsMonitorOnly` functions for `container` (the go struct that abstracts the `docker inspect` raw data about a container), we can add a `GlobalConfig` which includes the argument settings for `labelPrecendence`, `monitorOnly`, etc. 
That way we can keep the logic inside the `container` funcs,  and making this statement much more readable. What do you think? I can help with this during the weekend.

piksel: ...actually, we could just pass the `params` to the functions, since that is already basically the `GlobalConfig` 😅

jebabin: So i passed params to isMonitorOnly and the check is done inside.
(I initialy tried a `containerMonitorOnly` like varible before `shouldupdate` but it was even worse to read as the logic gets more complex)

PS: My test was wrong, I hope I fixed it correctly
PS2: I also fixed (I think) the previous test. At line 181, it was `_, err := actions.Update(client, types.UpdateParams{MonitorOnly: true})` and the check is `Expect(client.TestData.TriedToRemoveImageCount).To(Equal(0))` 
As there was no `Cleanup: true` in `UpdateParams`, for me there was no change that `TriedToRemoveImageCount` could be something else than 0

jebabin: As we discovered in #1753 that there is undocumented no-pull label, there is another change to do get expected behavior when label-take-precedence is used. Actual behavior is to not pull if either no-pull is set as argument or as label

jebabin: Actualy, no-pull label is documented in main, this option has been added in march

jebabin: I mark the conversation as resolved. label-take-precendence now also handle no-pull option

piksel:
I made a generalised helper to implement the logic, to make it easier to visually understand the logic/flow:

```go
func (c Container) getContainerOrGlobalBool(globalVal bool, label string, contPrecedence bool) bool {
	if contVal, err := c.getBoolLabelValue(label); err != nil {
		if !errors.Is(err, errorLabelNotFound) {
			logrus.WithField("error", err).WithField("label", label).Warn("Failed to parse label value")
		}
		return globalVal
	} else {
		if contPrecedence {
			return contVal
		} else {
			return contVal || globalVal
		}
	}
}
```

Are you OK with me pushing this? I can also add some more tests to increase the coverage.

jebabin:
Of course I'm, this is clearly better

piksel (approved):

PR #1755, opened 2023-09-08
chore: update Dockerfile.dev-self-contained to allow better build cache

jebabin:
Currently, when using the dockerfile, module are downloaded each time.
This change pre download the module in a early stage so that docker layer caching avoid module redownload at next build


0220tiezo1323 (changes requested):
No

piksel (approved):

PR #1767, opened 2023-09-13
docs: update interval text in introduction

valankar:
Update text to reflect the 24 hour default checking time.


piksel (approved):
That introduction text is very vague about the configuration used, but I guess it's better to assume that it's using the default schedule. Thanks!

PR #1783, opened 2023-10-03
Adding a flag/env var to explicitly exclude containers from updating

### LINKED ISSUE #1566: Non-label method for excluding containers

On Synology (and I assume other environments - UniFi comes to mind), there are often "private" containers that we don't control the creation of. As a consequence, watchtower tries to update those but fails.

Since there's no way to update labels for those containers after they were created, and we don't create them, there's no good way to exclude them, short of

I'd be happy with a `--exclude` or similar flag (possibly taking a regex?) to tell watchtower what to exclude - that's easy to update.

* Using `--label-enable` is harder to manage (requires managing labels for all the other containers)
* Creating derived containers just to add the label is not really an option for a lot of these private containers, especially as some of them have their own update mechanism.
* On Synology specifically, updating environment variables is easier (doesn't require recreating the container), so that's definitely a good option there
* There could simply be a flag to silently ignore any containers that aren't on the registry (would solve the Synology docviewer case, might not solve the UniFi case where they create common containers too)
* A file with the list of exclusions might also work

### Additional context

Related to https://github.com/containrrr/watchtower/issues/923 .

rdamazio:
This adds a way to exclude containers by name, rather than having to set labels.

This resolves #1566, where containers are managed by the OS and can't have labels applied easily.


rdamazio:
Hold on, I messed up my sync, I'll re-send.

PR #1784, opened 2023-10-03
feat(filters): Add a flag/env to explicitly exclude containers by name

### LINKED ISSUE #1566: Non-label method for excluding containers

rdamazio:
This adds a way to exclude containers by name, rather than having to set labels.

Resolves #1566

### DIFF
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -230,6 +230,19 @@ __Do not__ Monitor and update containers that have `com.centurylinklabs.watchtow
 no `--label-enable` argument is passed. Note that only one or the other (targeting by enable label) can be 
 used at the same time to target containers.
 
+## Filter by disabling specific container names
+Monitor and update containers whose names are not in a given set of names.
+
+This can be used to exclude specific containers, when setting labels is not an option.
+The listed containers will be excluded even if they have the enable filter set to true.
+
+```text
+            Argument: --disable-containers, -x
+Environment Variable: WATCHTOWER_DISABLE_CONTAINERS
+                Type: Comma- or space-separated string list
+             Default: ""
+```
+
 ## Without updating containers
 Will only monitor for new images, send notifications and invoke
 the [pre-check/post-check hooks](https://containrrr.dev/watchtower/lifecycle-hooks/), but will __not__ update the
--- a/internal/flags/flags.go
+++ b/internal/flags/flags.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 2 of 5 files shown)


internal/flags/flags.go
```diff
@@ -85,6 +86,13 @@ func RegisterSystemFlags(rootCmd *cobra.Command) {
 		envBool("WATCHTOWER_LABEL_ENABLE"),
 		"Watch containers where the com.centurylinklabs.watchtower.enable label is true")
 
+	flags.StringSliceP(
+		"disable-containers",
+		"x",
+		// Due to issue spf13/viper#380, can't use viper.GetStringSlice:
+		regexp.MustCompile("[, ]+").Split(viper.GetString("WATCHTOWER_DISABLE_CONTAINERS"), -1),
```
piksel: Use `envString` which makes sure the docs match the implementation:
```suggestion
		regexp.MustCompile("[, ]+").Split(envString("WATCHTOWER_DISABLE_CONTAINERS"), -1),
```

docs/arguments.md
```diff
@@ -230,6 +230,19 @@ __Do not__ Monitor and update containers that have `com.centurylinklabs.watchtow
...
+## Filter by disabling specific container names
+Monitor and update containers whose names are not in a given set of names.
+
+This can be used to exclude specific containers, when setting labels is not an option.
+The listed containers will be excluded even if they have the enable filter set to true.
+
+```text
+            Argument: --disable-containers, -x
+Environment Variable: WATCHTOWER_DISABLE_CONTAINERS
+                Type: Comma-separated string list
```
piksel: This can also be space-separated. Perhaps that should be noted as well (if it was intended).

piksel (changes requested):
Looks good! Just a two small requests...

docs/arguments.md
rdamazio: It was kind of intended to be consistent with viper's standard, but on a second thought there's no reason to, updating regex instead.

rdamazio: Done - do you think it'd make sense to separately just make envStringSlice always accept commas?

internal/flags/flags.go
rdamazio: Done

piksel (approved):

PR #1791, opened 2023-10-06
adding log error massage #1290

### LINKED ISSUE #1290: Can you please add this to the post-update
The (still undocumented) feature is great for our purpose, but can you add the same functionality to the post-update hook (in the same manner, of course)?

I have added a script to the post-update to check the database-version. If they not match, the script returns exit 1. It would be great if the post-update hook could log a error- or warning-message, so the slack-notification gets a message (level warn) and we are alarmed that something went wrong.

https://github.com/containrrr/watchtower/blob/9e7dfe7be442d1259b6ed5223558f2201f1fd43b/internal/actions/update.go#L76-L80

gaurangkudale:
adding log error massage for fix #1290

### DIFF
--- a/internal/actions/update.go
+++ b/internal/actions/update.go
@@ -222,10 +222,12 @@ func restartStaleContainer(container types.Container, client container.Client, p
 
 	if !params.NoRestart {
 		if newContainerID, err := client.StartContainer(container); err != nil {
+			log.Info("Error occurred while starting the container.")
 			log.Error(err)
 			return err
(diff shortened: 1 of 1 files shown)


piksel:
Sorry, this not the requested functionality. It has to do with logging and handling the outcome of the `ExecutePostUpdateCommand`. But it looks like it's already implemented...

simonaronsson:
As this has already been fixed, I'm closing the PR.

PR #1792, opened 2023-10-07
chore: replace usages of ioutil

donuts-are-good:
ioutil is not used anymore, in favor of io or os. in this case, io is a drop in replacement for ioutil.

### DIFF
--- a/pkg/api/metrics/metrics_test.go
+++ b/pkg/api/metrics/metrics_test.go
@@ -2,18 +2,18 @@ package metrics_test
 
 import (
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"strings"
 	"testing"
 
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
 	"github.com/containrrr/watchtower/pkg/api"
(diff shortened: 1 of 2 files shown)


donuts-are-good:
I'm happy to help get this merged. Please advise what is needed.

piksel:
Hello! Sorry, I have been a bit busy. The complaint is about how one of the changed lines has no test coverage, but this just because it had no coverage before. The PR is about replacing a call to a deprecated function with the non-deprecated version, so adding tests for the functionality might be a bit out of scope. Ideally you could add a test that "proves" that the new function does the same as before, but since it's just a rename/package move, that shouldn't be necessary.

donuts-are-good:
> Hello! Sorry, I have been a bit busy. The complaint is about how one of the changed lines has no test coverage, but this just because it had no coverage before. The PR is about replacing a call to a deprecated function with the non-deprecated version, so adding tests for the functionality might be a bit out of scope. Ideally you could add a test that "proves" that the new function does the same as before, but since it's just a rename/package move, that shouldn't be necessary.

Awesome thanks! that's reassuring :) Is there anything else that's needed before this is merged or should I sit tight and wait?

donuts-are-good:
Thanks :)

Issue #1716, opened 2023-08-05
Log to file


lonix1:
# Is your feature request related to a problem?

Watchtower seems to log to `stdout` only, and thus to docker's built-in "json driver". The logs can be access like so:
- `$ docker logs -t $CONTAINER_NAME`
- `tail -f /var/lib/docker/containers/$CONTAINER_ID/$CONTAINER_ID-json.log`

When the container is recycled, the logs are lost. The workaround is to use another logging tool, but that is unnecessary complexity, especially for small deployments.

Also, in general, the logs are not in a friendly format - it's much simpler, especially in small deployments, to simply have the log in a file that can be easily manipulated. Not everyone uses large complex logging apps and servers, etc.

# Describe the solution you'd like

Many apps have an option to log to file as well.

It would be nice if the log could be sent to a file instead of (or in addition to) stdout. If enabled, it could be written to `/var/log/watchtower.log` by default.

Personally, I would mount it as `/var/log/watchtower.log:/var/log/watchtower.log` so that I could have it on the host, and use logrotate on it.

And of course if one sends a `USR1` signal to the app it should release and recreate the file (so the host can work on it, e.g. logrotate).

# Describe alternatives you've considered

An unnecessarily complicated environment that includes log aggregation and monitoring, and who knows what else. For small servers that is overkill and hard to manage.

# Additional context

Thank you for considering it!

PR #1813, opened 2023-10-26
feat: add log file support

jokresner:
This PR adds two new flags to enable logging to a file. See #1716


simonaronsson (changes requested):
I like this, and would be happy to merge it after one change: enabling logging to a file should still log to stdout as well.

jokresner:
Sorry for my late replay. I've added a multi writer to support both outputs.

IvanRibakov:
@simskij Could you please take another look at this PR?

PR #1839, opened 2023-11-12
Add a warning about using Watchtower in production

simonaronsson:
Add a warning about using Watchtower in production, and a couple of recommended alternatives.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -6,6 +6,30 @@
   "commit": false,
   "contributors": [
     {
+      "login": "piksel",
+      "name": "nils måsén",
+      "avatar_url": "https://avatars2.githubusercontent.com/u/807383?v=4",
+      "profile": "https://piksel.se",
+      "contributions": [
+        "code",
+        "doc",
+        "maintenance",
+        "review"
+      ]
+    },
+    {
+      "login": "simskij",
+      "name": "Simon Aronsson",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/1596025?v=4",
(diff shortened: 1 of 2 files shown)


waja:
How about also declare that also on https://containrrr.dev/watchtower/?
Beside that, is there a special reason for this warning?

simonaronsson:
> How about also declare that also on https://containrrr.dev/watchtower/? Beside that, is there a special reason for this warning?

That will end up on the docs next time we cut a release. The reason is mainly setting expectations. Blindly updating containers is usually not a good practice in a setting where data consistency and reliability is required.

Historically, we've avoided making a stance either way, but in the light of the latest update spawning clones of watchtower itself, I felt a warning was in place so that users can make an informed, conscious choice about what risk levels they are willing to tolerate.

waja:
Yeah ... but everybody following `latest` (for example) in any way might be on the risky side and this is not an issue of watchtower but on using a moving target.
We are using watchtower for the most time to follow a "release"-tag ... for example for mariadb you can follow `lts` or what ever [tag](https://hub.docker.com/_/mariadb) you trust to follow. We use the x.y tags in this case. :)

piksel:
It still may be the case that the update leaves your environment broken. Watchtower tries its best to recreate containers using the same configuration, but there is always a possibility that the update fails and you're left without a running container (since we have to remove the old container before creating the new one). I do however agree with the notion of not telling people how they *should* (not) be using watchtower, but rather inform about what the consequences may be. Perhaps encouraging that you can create your containers from scratch for example, and keeping the valuable state in volumes etc.

PR #1841, opened 2023-11-12
document how to skip empty notifications using templates

### LINKED ISSUE #1200: Document how to avoid empty notifications when using notification templates
> It's the template that is causing this. It's an unintended side effect of allowing the template to dictate whether to send a notifiction. Fix for the authors template:
> 
> ```yaml
> WATCHTOWER_NOTIFICATION_TEMPLATE: "{{if .}}<b>Watchtower@hetzner</b>{{println}}{{end}}{{range .}}{{.Time.Format \"2006-01-02 15:04:05\"}} ({{.Level}}): {{.Message}}{{println}}{{end}}"
> ```

This is an excellent item to add to the "things to document"-backlog.

_Originally posted by @simskij in https://github.com/containrrr/watchtower/issues/1198#issuecomment-1019517527_

simonaronsson:
resolves #1200


piksel:
Hm, but that isn't entirely correct as it only applies to non-report templates.

(and the sentence before it only applies to report templates).

The difference being that the non-report templates only has a list of log messages as it's data.

PR #1895, opened 2024-01-06
refactor(update): clean up actions/update

piksel:
This PR tries to refactor the update code flow to make it easier to understand, maintain and extend.

actions/update:
 - move common arguments to a shared struct (`updateSession`)
 - remove unused fields
 - fix outdated names
 - improve logging/error handling

lifecycle:
 - removes unwieldy SkipUpdate return value in favor of errors.Is
 - generalizes the code for all four phases
 - allows timeout to be defined for all phases
 - enables explicit unit in timeout label values (in addition to implicit minutes)

container:
- rename `Stale` to `MarkedForUpdate`
renames the `container.Stale` field to what it's actually used for, as staleness
is not the only factor used to decide whether a container should be updated anymore

### DIFF
--- a/internal/actions/update.go
+++ b/internal/actions/update.go
@@ -2,6 +2,7 @@ package actions
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/containrrr/watchtower/internal/util"
 	"github.com/containrrr/watchtower/pkg/container"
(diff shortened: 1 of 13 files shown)


pkg/container/container.go
```diff
@@ -208,7 +208,7 @@ func (c Container) Links() []string {
 // ToRestart return whether the container should be restarted, either because
 // is stale or linked to another stale container.
```
pjdubya, line 209: The comment describing the function is out of date now with the revised logic

internal/actions/update.go
```diff
@@ -51,128 +67,122 @@ func Update(client container.Client, params types.UpdateParams) (types.Report, e
...
-	return
+	us.progress.UpdateFailed(failed)
+
+	return stopped
 }
 
-func stopStaleContainer(container types.Container, client container.Client, params types.UpdateParams) error {
-	if container.IsWatchtower() {
-		log.Debugf("This is the watchtower container %s", container.Name())
+func (us *updateSession) stopContainer(c types.Container) error {
```
pjdubya, line 146: Consider keeping this variable named "container" rather than "c" to help in readability and to match other functions like restartContainer

internal/actions/update.go
```diff
@@ -181,58 +191,58 @@ func restartContainersInSortedOrder(containers []types.Container, client contain
 			continue
 		}
 		if stoppedImages[c.SafeImageID()] {
-			if err := restartStaleContainer(c, client, params); err != nil {
+			if err := us.restartContainer(c); err != nil {
 				failed[c.ID()] = err
-			} else if c.IsStale() {
+			} else if c.IsMarkedForUpdate() {
 				// Only add (previously) stale containers' images to cleanup
 				cleanupImageIDs[c.ImageID()] = true
```
pjdubya, line 198: To me, rather than at line 89, this is a better place to call

			us.progress.MarkForUpdate(c.ID())

That would avoid setting a container's state to UpdatedState before it's actually updated. That would also avoid an item going from UpdatedState to FailedSate in cases where something fails (or if something later breaks with properly setting FailedState, misrepresenting the item as having been updated if it wasn't). This is probably not a big deal but I mention it because it was confusing to me initially why we're setting something to (past-tense) UpdatedState before any action is completed.

internal/actions/update.go
```diff
@@ -51,128 +67,122 @@ func Update(client container.Client, params types.UpdateParams) (types.Report, e
...
-		return nil, err
+		return fmt.Errorf("failed to sort containers for updating: %v", err)
 	}
 
 	UpdateImplicitRestart(containers)
 
 	var containersToUpdate []types.Container
 	for _, c := range containers {
-		if !c.IsMonitorOnly(params) {
+		if c.ToRestart() {
```
pjdubya, line 87: In testing my latest changes that assume this PR will be coming in, I found that I needed to change this:

		if c.ToRestart() {

to

		if c.ToRestart() && !c.IsMonitorOnly(params) {

In order to avoid the monitor only test cases failing. Consider whether to make the same change here in your code or just to incorporate the  	!c.IsMonitorOnly(params)  test within the definition of ToRestart()

internal/actions/update.go
piksel: Can't, since that is the name of an imported package. This is the common go "solution". :/

internal/actions/update.go
piksel: It's not needed anymore, since `shouldUpdate` (and later `markedForUpdate`) will never be true for `monitorOnly`.

pjdubya:
Hi @piksel . Just curious if there was any movement on this PR? Would love to update my other PR for deferred updates to match this code and get that resolved too whenever you're ready. LMK if there's anything else I can do to help here!

minscof:
It would be nice to finish this work

MattBlackOnly:
@piksel  would also be interested in this too for defer functionality

PR #1924, opened 2024-02-06
Add OCI image labels

Qup42:
This PR adds Labels to the Docker Image. The added Labels conform to the [Annotations defined in the OCI Image Format Specification](https://github.com/opencontainers/image-spec/blob/main/annotations.md).
The following labels are added for releases (dev and tagged):
- `org.opencontainers.image.url`
- `org.opencontainers.image.documentation`
- `org.opencontainers.image.source`
- `org.opencontainers.image.licenses`
- `org.opencontainers.image.title`
- `org.opencontainers.image.description`
- `org.opencontainers.image.base.name`

Tagged releases are additionally labeled with
- `org.opencontainers.image.created`
- `org.opencontainers.image.version`
- `org.opencontainers.image.revision`

Image Labels are considered a best practice for Container Images. They make it easier to handle Images and reconstruct which version one is running. Image Labels also enable tools for automatic dependency updates (like Renovate and Dependabot) to automatically retrieve and display Release Notes for an update.

I tested the changes locally with `goreleaser release --skip-publish --skip-validate --rm-dist`. The images had the expected labels. There are no existing test suites that apply to this change.

Documentation updates are not required for this change.

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -12,7 +12,7 @@ LABEL "org.opencontainers.image.url"="https://containrrr.dev/watchtower/" \
       "org.opencontainers.image.licenses"="Apache-2.0" \
 # unsure about capitalization
       "org.opencontainers.image.title"="watchtower" \
-      "org.opencontainers.image.title"="A process for automating Docker container base image updates." \
+      "org.opencontainers.image.description"="A process for automating Docker container base image updates." \
 # The version might be forgotten. The version could be extraced into a `ARG` before the initial `FROM` or removed.
       "org.opencontainers.image.base.name"="alpine:3.19.0"
 
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -6,6 +6,15 @@ RUN apk add --no-cache \
 
 FROM scratch
 LABEL "com.centurylinklabs.watchtower"="true"
+LABEL "org.opencontainers.image.url"="https://containrrr.dev/watchtower/" \
+      "org.opencontainers.image.documentation"="https://containrrr.dev/watchtower/" \
+      "org.opencontainers.image.source"="https://github.com/containrrr/watchtower" \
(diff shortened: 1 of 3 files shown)


dockerfiles/Dockerfile
```diff
@@ -6,6 +6,15 @@ RUN apk add --no-cache \
 
 FROM scratch
 LABEL "com.centurylinklabs.watchtower"="true"
+LABEL "org.opencontainers.image.url"="https://containrrr.dev/watchtower/" \
+      "org.opencontainers.image.documentation"="https://containrrr.dev/watchtower/" \
+      "org.opencontainers.image.source"="https://github.com/containrrr/watchtower" \
+      "org.opencontainers.image.licenses"="Apache-2.0" \
+# unsure about capitalization
+      "org.opencontainers.image.title"="watchtower" \
```
Qup42, line 14: Are there any feelings about the capitalization of watchtower? As a non-native speaker I am a bit unsure.

dockerfiles/Dockerfile
```diff
@@ -6,6 +6,15 @@ RUN apk add --no-cache \
...
+      "org.opencontainers.image.description"="A process for automating Docker container base image updates." \
+# The version might be forgotten. The version could be extraced into a `ARG` before the initial `FROM` or removed.
+      "org.opencontainers.image.base.name"="alpine:3.19.0"
```
Qup42, line 17: Any feelings? If not I would tend towards removing the concrete version. It adds little in additional information but makes the Dockerfile somewhat more complex.

tko: You could be able to do something like
```Dockerfile
ARG BASE_IMAGE=alpine:3.19.0

FROM --platform=$BUILDPLATFORM $BASE_IMAGE as alpine
[..]
FROM scratch
ARG BASE_IMAGE
LABEL "org.opencontainers.image.base.name"="$BASE_IMAGE"
```

can't comment on if it would be useful in practice (can't recall ever looking for the labels myself)

PR #1978, opened 2024-05-08
feat: add ability to define original image id and image name used in custom images 

aminfa:
TL;DR: This PR allows one to add two labels to containers that define which original image was used as the base of the image being run. 

## The problem

Sometimes it is necessary to build a custom docker image that only slightly changes a base image. For example:

```
FROM nginx:1.25.4
COPY nginx.conf /etc/nginx/nginx.conf
```

When running this custom image, watchtower cannot figure out that the base nginx image is stale, as it instead only checks if our custom nginx image is stale:

```
Could not do a head request for "custom-nginx:latest", falling back to regular pull.
Reason: registry responded to head request with "401 Unauthorized", auth: "Bearer realm=\"https://auth.docker.io/token\",service=\"registry.docker.io\",scope=\"repository:library/nginx-as-container-rootnginx:pull\",error=\"insufficient_scope\""
Unable to update container "custom-nginx-1": Error response from daemon: pull access denied for custom-nginx, repository does not exist or may require 'docker login': denied: requested access to the resource is denied. Proceeding to next.
```

## The solution

With this PR, it is possible to add two labels to the container that overwrites the image ID and image name used by watchtower to decide whether the image is stale:

```
docker run -d \
   -l com.centurylinklabs.watchtower.original-image=nginx \
   -l com.centurylinklabs.watchtower.original-image-id=$(docker inspect nginx:1.25.4 | jq '.[0]'.Id) \
   -l com.centurylinklabs.watchtower.monitor-only=1 \
    custom_nginx
```

Now, watchtower reports a new nginx version:

```
$ ./watchtower --run-once
INFO[0001] Watchtower v0.0.0-unknown                    
INFO[0001] Using no notifications                       
INFO[0001] Checking all containers (except explicitly disabled with label) 
INFO[0001] Running a one time update.                   
INFO[0003] Original image id for nginx:latest found: ("sha256:c613f16b6642) 
INFO[0003] Found new nginx:latest image (1d668e06f1e5)
```

## Current limitations

Currently automatic updating of the images will break the containers as it will replace the container using images of the original base image. Here it would start `nginx:1.25.5` instead of a custom version of it. 
Thus, it is necessary to set `monitor-only=1`.

## Further work

I haven't done: "Tests that verify the code your contributing" and "Updates to the documentation" because I wanted to wait and see if this feature is even desired. 

Feel free to reject the PR if the feature does not fit within the scope of this project.

### DIFF
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -329,10 +329,12 @@ func (client dockerClient) IsContainerStale(container t.Container, params t.Upda
 }
 
 func (client dockerClient) HasNewImage(ctx context.Context, container t.Container) (hasNew bool, latestImage t.ImageID, err error) {
-	currentImageID := t.ImageID(container.ContainerInfo().ContainerJSONBase.Image)
+	container_info := container.ContainerInfo()
+	currentImageID := t.ImageID(container_info.ContainerJSONBase.Image)
 	imageName := container.ImageName()
 
-	imageIDSetByLabel, ok := container.ContainerInfo().Config.Labels[originalImageIDLabel]
+	// If the original-image-id label is set, it overwrites the image id reported by docker
+	imageIDSetByLabel, ok := container_info.Config.Labels[originalImageIDLabel]
 	if ok {
 		currentImageID = t.ImageID(imageIDSetByLabel)
 		log.Infof("Original image id for %s found: (%s)", imageName, currentImageID.ShortID())
--- a/pkg/container/client.go
+++ b/pkg/container/client.go
@@ -332,6 +332,12 @@ func (client dockerClient) HasNewImage(ctx context.Context, container t.Containe
 	currentImageID := t.ImageID(container.ContainerInfo().ContainerJSONBase.Image)
 	imageName := container.ImageName()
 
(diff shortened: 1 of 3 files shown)


kondas:
+1, this is a helpful feature for myself, and others who are using watchtower for monitoring only, and have custom local built images.

One very useful addition would be the possibility to have the ID of the base image optional. I dont really know if its a workable approach, but would perfectly fit my use case, when using fixed version tags in the base images, not rolling ones. In this case watchtower should gather the ID for the base image found in com.centurylinklabs.watchtower.original-image label.

This way the addition of the base image is straightforward in the local build context like this:

```
ARG RUNDECK_IMAGE
FROM ${RUNDECK_IMAGE:-rundeck/rundeck:SNAPSHOT}
<stuff>
ARG RUNDECK_IMAGE
LABEL com.centurylinklabs.watchtower.original-image=${RUNDECK_IMAGE:-rundeck/rundeck:SNAPSHOT}
```
docker-compose.yml
```
services:
  app:
    pull_policy: build
    build:
      context: ./build/
      args:
        RUNDECK_IMAGE: rundeck/rundeck:5.4.0
```

PR #2054, opened 2024-11-21
add optional image tag filter

### LINKED ISSUE #1932: Update/Downgrade an container to a specific tag via http api mode

I'd like to update **OR** downgrade a specific container via http mode.
I'm using `` localhost:8080/v1/update?images=<image_name>`` to update my containers via http.
However, in case the new container contain any errors I'd like to downgrade them again to the tag that was used before.

Allow to add an image name with a **certain tag**.
In this way we could define which tag should be used.

**Example:**
`` localhost:8080/v1/update?images=<image_name>:<tag>``

Another possibility would be an api call "downgrade". The approach, however, would be less flexible as it would only reset the container to the penultimate tag.

**Example:**
`` localhost:8080/v1/downgrade?images=<image_name>``

### Additional context

_No response_

BenjaminFaal:
To be able to do finegrained updates to container running a specific tag of an image i added this.

closes #1932

### DIFF
--- a/docs/http-api-mode.md
+++ b/docs/http-api-mode.md
@@ -43,3 +43,8 @@ In order to update only certain images, the image names can be provided as URL q
 ```bash
 curl -H "Authorization: Bearer mytoken" localhost:8080/v1/update?image=foo/bar,foo/baz
 ```
+
+tags are also supported:
+```bash
+curl -H "Authorization: Bearer mytoken" localhost:8080/v1/update?image=foo/baz:latest
+```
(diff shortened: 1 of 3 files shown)


Foxite:
I originally wrote the code for the ?image parameter in #1289 and made it ignore the specific tag. To be honest, I forgot why I did it like that, and it's now causing problems for me.

What's the status on this? It looks good to me.

Foxite (approved):

BenjaminFaal:
I am using this for a few months in production and it works.

Foxite:
I noticed this project seems dead; I'm examining forks by [nicholas-fedor](https://github.com/nicholas-fedor/watchtower) and [beatkind](https://github.com/beatkind/watchtower) to see if this can be incorporated there.

Foxite:
@BenjaminFaal are you hosting your fork image somewhere? If so, I think I'll just use yours.

PR #2084, opened 2025-02-20
Fix Unsafe Command Expansion in build-tplprev.sh

advaitpatel:
The script uses `cd $(git rev-parse --show-toplevel)`, which fails if the path has spaces.

Therefore, replaced `cd $(command)` with `cd "$(command)"` to properly handle spaces.

After these changes. it prevents build failures due to incorrect path handling and improves error reporting too.


advaitpatel:
Hello @simskij / @piksel can i please get a review on this?

